PR #1, opened 2022-12-16
feat: init repo

FoseFx:
I realize this a massive PR, so here are some pointers (also I recommend to check it out locally):
there are `/libs` and `/services`, the latter only has the new `user-svc` service.

In `libs` there are
*  `common` - the library for everything related to setup and dapr
* `logging` - zerolog helpers
* `hwgorm` - gorm helpers
* `hwutil` - general purpose utility library


JannikWibker:
image

any particular reason for this?, this seems like a rather long list

FoseFx:
copy & paste from hw/rest-api

JannikWibker:
ahh true
then the question becomes why hw/rest-api had to have gitignore entries for vim, emacs, dart, flutter, android, pycharm, react, react native, python, xcode, node, cordova, vue, etc.

MaxSchaefer:
The user-svc is a nice example for now, but in the futures the organization-part should be handled by a separate service.

FoseFx:
As of today's meeting we scrap Auth v1 anyway (in favor of a IAM solution, Keycloak?), so this service will probably experience heavy changes anyway.

MaxSchaefer:
> As per today's meeting we scrap Auth v1 anyway (in favor of a IAM solution, Keycloak?), so this service will probably experience heavy changes anyway.

Yes, Keycloak was the plan.

.gitignore
```diff
@@ -0,0 +1,1569 @@
+# Created by https://www.toptal.com/developers/gitignore/api/vim,emacs,dart,flutter,android,androidstudio,intellij,intellij+all,intellij+iml,pycharm,pycharm+all,pycharm+iml,go,goland,goland+iml,goland+all,visualstudiocode,react,reactnative,python,pythonvanilla,xcode,xcodeinjection,node,nativescript,cordova,apachecordova,ionic3,vue,vuejs
```
MaxSchaefer, line 1: Already discussed [here](https://github.com/helpwave/services/pull/1#issuecomment-1356375726).

services/user-svc/README.md
```diff
@@ -0,0 +1,4 @@
+# user-svc
+
+> User and Organization management service
```
MaxSchaefer, line 3:
```suggestion
> User management service
```

MaxSchaefer (changes requested):

.gitignore
```diff
@@ -0,0 +1,481 @@
+# Created by https://www.toptal.com/developers/gitignore/api/go,vim,goland,intellij,goland+all,goland+iml,intellij+all,intellij+iml,visualstudiocode
```
MaxSchaefer, line 1: I guess not perfect but okay for now.

MaxSchaefer (approved):
Nice! This should unblock the other pull requests.

PR #24, opened 2022-12-20
feat: build docker from scratch

use-to:
This change will make our docker images smaller!

### DIFF
--- a/services/user-svc/Dockerfile
+++ b/services/user-svc/Dockerfile
@@ -6,7 +6,7 @@ COPY . /app
 ARG VERSION
 RUN CGO_ENABLED=0 GOOS=linux go build -ldflags="-X main.Version=${VERSION}" -a -o app .
 
-FROM alpine AS production
+FROM scratch AS production
 
 WORKDIR /app
 
@@ -21,4 +21,3 @@ ENV POSTGRES_PORT 5432
 EXPOSE 80
 
 CMD ["./app"]
-


use-to:
There is still a problem with this docker build! See #25

JannikWibker:
Is using a scratch image a good idea?

When it comes to docker images with a fairly small size there are things such as [distroless images](https://github.com/GoogleContainerTools/distroless) which I've previously used and had success with

FoseFx:
I think scratch images are missing CAs so we can not do secure https requests anymore. (And there's probably a lot of other pain points that we will run into using scratch).

IIRC there are other "minimal" scratch-based images that have some of these features. busybox comes to mind, but I'm not sure what features it has and what is missing.

MaxSchaefer:
We can just copy the CAs from our builder like `COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/`. If this should be one of the few problems with scratch, let's try to be as minimal as possible by using scratch.

FoseFx:
Turns out CAs are not the only thig missing, I found [this article](https://chemidy.medium.com/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324), which does into a bit more detail

use-to:
It's just a binary we can ignore those "side-effects".
But I don't think its worth the discussion.
So let's stick with the current `alpine` version!

PR #31, opened 2022-12-27
build: add all target

MaxSchaefer:


Makefile
```diff
@@ -7,6 +7,10 @@ SERVICES = user-svc emergency-room-svc
 $(SERVICES):
 	docker build -f ${DOCKERFILE_SERVICES} --build-arg=VERSION=${VERSION} --build-arg=SERVICE=$@ -t helpwave/$@ .
 
+.PHONY: all
+all:
+	$(foreach SERVICE, ${SERVICES}, $(MAKE) ${SERVICE})
```
FoseFx, line 12: this will generate the command `make user-svc  make emergency-room-svc`, so make will try to build `user-svc`, `emergency-room-svc` but also the target `make`, which does not exist.

> make[1]: *** Keine Regel, um „make“ zu erstellen.  Schluss.

FoseFx (changes requested):

Makefile
FoseFx: we can simply use all services as dependencies for an `all` target like this:
```make
.PHONY: all
all: $(SERVICES)
        echo "built all"
```

MaxSchaefer: Yeah, seams like a nice solution. This issue was not really visible in the output, `make -n all` really helps :D .

FoseFx (approved):

PR #32, opened 2022-12-29
Add ci for our monorepo setup

### LINKED ISSUE #30: Building images in GitHub Actions
We should build our images as we have already done it in helpwave/rest-api. In the case of helpwave/services, we need to build images when changes occur in the /services directory. We can use the base name of the directory that changed as our image tag. We should not build all images when only one directory in /services has changes. **This issue should only cover actions that runs on *push* (branch and pull-requests). Tagging and deployment should be handled separately.**

For example:
- A change happens in /services/user-svc
- The change gets pushed
- Linting runs globally
- Migration testing and building runs in context of the /services/user-svc directory

MaxSchaefer:
Most of `.github/workflows/cicd-go.yaml` is copied over from `helpwave/rest-api` and works as a template for services written in go. `.github/workflows/cicd.yaml` detects and aggregates changes in libs or services and triggers all or specific ci`s.

Closes #30

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,7 @@ insert_final_newline = true
 charset = utf-8
 indent_style = tab
 indent_size = 4
+
+[*.yaml]
+indent_style = space
+indent_size = 2
--- a/.github/workflows/cicd-go.yaml
+++ b/.github/workflows/cicd-go.yaml
@@ -1,0 +1,150 @@
+name: 'CICD Go'
+
+on:
+  workflow_call:
+    inputs:
+      service:
+        description: 'The name of the service.'
+        type: string
+        required: true
+
+permissions:
+  contents: read
+  pull-requests: read
+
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -10,3 +10,6 @@ insert_final_newline = true
 charset = utf-8
 indent_style = tab
 indent_size = 4
+
+[*.yaml]
+indent_size = 2
```
FoseFx, line 15: while we are at it, also add `indent_style = spaces` _[see why](https://yaml.org/faq.html)_

FoseFx (approved):
nice solution! have you tried it out somewhere already?

.editorconfig
MaxSchaefer: > nice solution! have you tried it out in another branch/repo already?

Yes, extensively. Take a look at https://github.com/helpwave/helpwave-maxrs-services *(cloned from helpwave/services into my personal account, moved to our organization for this pr. Gets deleted when this pr gets merged.)*

PR #46, opened 2023-01-20
feat(er-svc): add update route

### LINKED ISSUE #6: Update ER

FoseFx:
Closes #6

### DIFF
--- a/services/emergency-room-svc/main.go
+++ b/services/emergency-room-svc/main.go
@@ -36,6 +36,7 @@ func main() {
 	common.MustAddServiceInvocationHandler(service, "create-emergency-room", createERHandler)
 	common.MustAddServiceInvocationHandler(service, "get-emergency-room", getERHandler)
 	common.MustAddServiceInvocationHandler(service, "get-emergency-rooms", getERsHandler)
+	common.MustAddServiceInvocationHandler(service, "update-emergency-room", updateERHandler)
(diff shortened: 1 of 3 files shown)


services/emergency-room-svc/main.go
```diff
@@ -216,6 +217,85 @@ func whereClausesForERsQuery(db *gorm.DB, request *api.GetERsRequestV1) *gorm.DB
...
+	log.Debug().Str("body", logging.Formatted(request)).Send()
+
+	// Update
+	db := hwgorm.GetDB(logCtx)
+
+	emergencyRoom := models.EmergencyRoom{
+		ID: request.ID,
+	}
+
+	updateMap := updatesMap(&request)
```
MaxSchaefer, line 240: Can we "map" the UpdateERRequestV1 struct to an EmergencyRoom struct and use this for the update instead of a map?

MaxSchaefer (changes requested):

services/emergency-room-svc/main.go
FoseFx: no because of go's handling of default values. e.g. updating open to "false" will not do anything as gorm thinks it's a default value and ignore it

MaxSchaefer: Okay, i guess then we should add the updatesMap function as a method to the UpdateERRequestV1 struct.

MaxSchaefer (approved):

PR #59, opened 2023-02-05
issue/50 organizations

### LINKED ISSUE #50: Create new Organization
As per https://github.com/helpwave/helpwave/issues/13:
> An organization inside helpwave is the hierarchical super node. An organization should represent exactly one hospital. An organization should have very basic information like a name.

Implement organizations as a combination of keycloak groups (e.g., `orga-b391150b-fb12-4870-9158-c5212f63319f`).

An organization must have an "is_personal" attribute, which can be true or false (default: false).

When creating an organization group it must have a sub-group called "admins", and add the creating user by default.

The semantics of these are described in https://github.com/helpwave/helpwave/issues/13.

### Additional notes
I'm not sure if a user can be in a sub-group and the parent group at the same time, if so we add the user to both the admin group AND the orga group. This makes promoting and demoting users easier and less fragile.

FoseFx:
Closes #50

### DIFF
--- a/libs/common/keycloak.go
+++ b/libs/common/keycloak.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"github.com/coreos/go-oidc"
 	"golang.org/x/oauth2"
 	"hwutil"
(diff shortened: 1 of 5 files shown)


libs/common/keycloak.go
```diff
@@ -17,11 +18,21 @@ import (
...
 type AccessTokenClaims struct {
-	// TODO
+	// Sub is the user id
+	Sub        string `json:"sub"`
+	GivenName  string `json:"given_name"`
+	FamilyName string `json:"family_name"`
+	Email      string `json:"email"`
 }
 
 func (a AccessTokenClaims) AsExpected() error {
```
MaxSchaefer, line 28: Right now, you create a struct of `AccessTokenClaims`, set the values and run `.AsExpected()`. Do you think it makes sense to handle these kinds of validation inside a constructor? Something like `NewAccessTokenClaims` where you pass `Sub` and `Email`, test against your cases and return a new struct of `AccessTokenClaims`.

MaxSchaefer (changes requested):

libs/common/keycloak.go
FoseFx: `AccessTokenClaims` should only be constructed in [VerifyAccessToken](https://github.com/helpwave/services/blob/4c705e8e1d3c10755570719586da10d71a6c3c98/libs/common/keycloak.go#L75) and filled by the token library using a reference. So this is not possible.

MaxSchaefer (approved):

PR #71, opened 2023-02-10
refactor: embrace terraform

### LINKED ISSUE #69: Ditch helmfiles, embrace terraform
`/kubernetes` is a hot mess right now. We use both tf and helmfiles.

Replace helmfiles with `.tf` files.

FoseFx:
Closes #69 (nice)


infrastructure/terraform/modules/hw-tf-module-cluster-resources/user-svc.tf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+
+resource "helm_release" "user-svc" {
+  name  = "user-svc"
+  chart = "../../../charts/service"
+  wait = true
+
+  depends_on = [
+    // helm_release.apisix,
```
MaxSchaefer, line 14:
```suggestion
    helm_release.dapr,
```

infrastructure/terraform/modules/hw-tf-module-cluster-resources/auth-svc.tf
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+resource "helm_release" "auth-svc" {
+  name  = "auth-svc"
+  chart = "../../../charts/service"
+  wait = true
+
+  depends_on = [
+    // helm_release.apisix,
```
MaxSchaefer, line 15:
```suggestion
    helm_release.dapr,
```

infrastructure/terraform/modules/hw-tf-module-cluster-resources/emergency-room-svc.tf
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+
+resource "helm_release" "emergency-room-svc" {
+  name  = "emergency-room-svc"
+  chart = "../../../charts/service"
+  wait = true
+
+  depends_on = [
+    // helm_release.apisix,
```
MaxSchaefer, line 50:
```suggestion
    helm_release.dapr,
```

infrastructure/terraform/modules/hw-tf-module-cluster-resources/keycloak.tf
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// keycloak operator + CRD + Ingress
+resource "helm_release" "keycloak" {
+  name  = "keycloak"
+  chart = "../../../charts/keycloak"
+  wait = true
+
+  depends_on = [
+    helm_release.apisix,
```
MaxSchaefer, line 86: Why does Keycloak depend on APISIX?

```suggestion
```

MaxSchaefer (changes requested):
Thanks for the clean-up 🧹

I've pushed a commit to use the `postgres` user instead of the `helpwave` user for the database on the emergency-room-svc. We need to dig into this later, we had the same issue before the refactor.

Every service that depends on keycloak enters an "CrashLoopBackOff" with the following error:
```bash
{"level":"fatal","version":"90398c0","service":"auth-svc","error":"Get \"https://keycloak-service.keycloak.svc.cluster.local:8443/realms/hw-staging/.well-known/openid-configuration\": x509: certificate is not valid for any names, but wanted to match keycloak-service.keycloak.svc.cluster.local","time":1676142059,"caller":"/libs/common/keycloak.go:143"}
```

infrastructure/terraform/modules/hw-tf-module-cluster-resources/keycloak.tf
FoseFx: Not now, but later this chart will deploy a CRD for routing. But for that the CRD definition needs to exist. Hence the dependence.

MaxSchaefer:
Looks good. What's up with the suggested changes? @FoseFx

MaxSchaefer:
Every service that depends on Keycloak, throws this new error:
```
{"level":"fatal","version":"90398c0","service":"auth-svc","error":"Get \"https://keycloak-service.keycloak.svc.cluster.local:8443/realms/hw-staging/.well-known/openid-configuration\": dial tcp 10.102.55.94:8443: connect: connection refused","time":1676409546,"caller":"/libs/common/keycloak.go:143"}
```
I guess that error gets thrown because Keycloak is not yet configured. How do I access the Admin UI? A port-forward seams not to work because of the Hostname and Hostport configuration.

FoseFx:
I made some changes. Now there is a DNS rewrite of `sso.helpwave.de` to `keycloak-internal.keycloak.svc.cluster.local` within the cluster. That's just a service in front of keycloak.

For local TLS verification can simply be disabled, for staging this is a very bad idea. Thus I outlined the cert-manager changes needed. Can you finish this as I don't have access to the cluster? You just have to get the cert into a tlsSecret that kc can accept.

This whole thing is needed, as OIDC requires the issuer to always match the requested base url. I tried to hack around it before and it just caused more breakage. I think the DNS rewrite is the best solution for this.

infrastructure/terraform/environments/staging/coredns.tf
```diff
@@ -0,0 +1,91 @@
...
+resource "kubectl_manifest" "coredns_config_map" {
+  depends_on = [
+    helm_release.coredns
+  ]
+
+  yaml_body = <<YAML
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-coredns
```
FoseFx, line 37: from what I can see the only reason to deploy this is the different name (`coredns` vs `coredns-coredns`). Can we just use a variable for this in the cluster resources and set it differently for staging, or will this cause issues with GKE? Having the same config at two places will undoubtedly lead to inconsistencies.

infrastructure/terraform/modules/hw-tf-module-cluster-resources/apisix.tf
```diff
@@ -123,7 +123,7 @@ resource "helm_release" "apisix" {
 }
 
 # kubectl port-forward <name of the dashboard-pod> 9000:9000
-resource "helm_release" "apisix-dashboard" {
+/*resource "helm_release" "apisix-dashboard" {
```
FoseFx, line 126: left over from testing? (also see the other services)

infrastructure/terraform/environments/staging/coredns.tf
MaxSchaefer: This was a dependency issue. You cannot create the ConfigMap for CoreDNS before the completion of the release. Fixed!

MaxSchaefer (approved):

PR #74, opened 2023-02-13
feat(infra): setup APISIX to proxy gRPC traffic to Dapr

### LINKED ISSUE #66: Adopt the custom resources of APISX
Right now, when APISIX gets deployed via our IaC. The first manual step is to setup the upstream, services and routes via the dashboard of APISIX.

The problem mentioned above can be solved by using the custom resources in our IaC which are already installed with APISIX.

**Definition of done**
A gRPC requests against a service behind `staging-api.helpwave.de:443` should be successful.

Define upstream, service and route to the apisix-dapr sidecar

MaxSchaefer:
This pull-request closes the gap between our gRPC microservices, Dapr and APISIX. When the Terraform setup gets applied and the services are started, we can access them via gRPC through APISIX in the Dapr manner.

```
// Running in the /services/emergency-room-svc directory
> grpcurl -import-path ./api -proto ./api/emergency-room-svc.proto -H 'dapr-app-id: emergency-room.emergency-room' -d '{}' staging-api.helpwave.de:443 EmergencyRoomService.GetERs
{
  "pageInfo": {
    "page": 1,
    "pageSize": 10,
    "totalSize": "2"
  }
}
```

In the future, we should provide direct gRPC endpoints (without the need to set the dapr-app-id header) with strong authentication via OIDC.

### Why are we using the [kubectl](https://registry.terraform.io/providers/gavinbunney/kubectl/latest) instead of the [kubernetes](https://registry.terraform.io/providers/hashicorp/kubernetes/latest) Terraform provider?
The kubernetes Terraform provider has a dependency during the PLAN phase of Terraform `terraform plan` or `terraform apply` on a available and ready to use ClusterAPI of Kubernetes. I found an interesting [Medium article](https://medium.com/@danieljimgarcia/dont-use-the-terraform-kubernetes-manifest-resource-6c7ff4fe629a) which explains the topic in more detail. You can see the result of my research in that pull-request, switching to [kubectl](https://registry.terraform.io/providers/gavinbunney/kubectl/latest) as our Terraform provider to interact with resources inside our cluster. The [kubectl](https://registry.terraform.io/providers/gavinbunney/kubectl/latest) has no PLAN dependency on the ClusterAPI, it applies YAML directly and tracking the changes.

### Why bypassing Cloudflare and using cert-manager?
There are several requirements for tunneling gRPC through the Cloudflare proxy. The most important and obvious requirement is that our origin server (in our case *Google Cloud L4 LB* -> *APISIX*) must support TLS on Port 443 and SNI for the certificate that is issued by Cloudflare to establish the communication between the Cloudflare proxy and our origin server, also called ["Full - SSL/TLS encryption mode"](https://developers.cloudflare.com/ssl/origin-configuration/ssl-modes/full/). Sounds like a pretty usual setup.

Unfortunately, this triggers the following error in APISIX: `http_ssl_phase(): failed to fetch ssl config: failed to find SNI: please check if the client requests via IP or uses an outdated protocol. If you need to report an issue, provide a packet capture file of the TLS handshake.`. 
*During my testing and debugging, the correct hostname was also in the log. That means APISIX was aware of the Hostname. And yes, the issued origin certificate was also in APISIX registered, and yes, also in the etcd. I've also swapped out APISIX with nginx (plus the origin certificate for sure) in our cluster, works perfectly fine.*
After hours of debugging, I've decided to go forward and ditching Cloudflare for now. It should be clear that the issue is most likely on the APISIX side, not Cloudflares. **Big TODO. We should dig into this.**
Here are similar errors, which have already been addressed but not solved yet
- https://github.com/apache/apisix/issues/4394
- https://github.com/apache/apisix/issues/8207

**With this pull-request, the IP of our GC LB is public and gets not proxied by Cloudflare. The SSL certificate gets issued by Lets Encrypt.**

---
Closes #66


kubernetes/iac/environments/staging/cert.tf
```diff
@@ -0,0 +1,94 @@
...
+  issuerRef:
+    name: letsencrypt
+    kind: Issuer
+    group: cert-manager.io
+YAML
+}
+
+resource "kubectl_manifest" "cert_apisix_tls_x-helpwave-de" {
+  depends_on = [
+    module.cluster-resources
```
FoseFx, line 76: does this not depend on `cert_certificate_x-helpwave-de`?

kubernetes/iac/environments/staging/cert.tf
```diff
@@ -0,0 +1,94 @@
...
+kind: Certificate
+metadata:
+  name: x-helpwave-de
+  namespace: ${local.cert_manager_namespace}
+spec:
+  secretName: x-helpwave-de
+  dnsNames:
+    - ${cloudflare_record.api-helpwave-de.hostname}
+    - ${cloudflare_record.staging-helpwave-de.hostname}
+    - ${cloudflare_record.staging-api-helpwave-de.hostname}
```
FoseFx, line 66: `kc.hw.de` domains missing

kubernetes/iac/modules/hw-tf-module-cluster-resources/apisix.tf
```diff
@@ -85,3 +90,48 @@ resource "helm_release" "apisix-dashboard" {
...
+kind: ApisixRoute
+metadata:
+  name: ${local.apisix_name}-dapr
+  namespace: ${local.apisix_namespace}
+spec:
+  http:
+  - backends:
+    - serviceName: ${local.apisix_name}-dapr
+      servicePort: 50001
+    match:
```
FoseFx, line 132: we should restrict this to only the api hosts (`api.hw.de`, `staging-api.hw.de`). this way we can add routes for other hosts (like `kc.hw.de`)

kubernetes/iac/modules/hw-tf-module-cluster-resources/apisix.tf
```diff
@@ -85,3 +90,48 @@ resource "helm_release" "apisix-dashboard" {
...
+  ]
+
+  yaml_body = <<YAML
+apiVersion: apisix.apache.org/v2
+kind: ApisixRoute
+metadata:
+  name: ${local.apisix_name}-dapr
+  namespace: ${local.apisix_namespace}
+spec:
+  http:
```
FoseFx, line 128: I don't really like this solution as it exposes the "dapr way" to the outside. It's better than having nothing running for now, but the API should be a gRPC (Web) API, not a dapr API.

kubernetes/iac/environments/staging/cert.tf
MaxSchaefer: Yes, but no 😄 APISIX tries to re-apply the tls resource until the secret is available. If we want to depend on it, we cannot do this with the tf `depends_on`.

kubernetes/iac/environments/staging/cert.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  yaml_body = <<YAML
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cloudflare-api-token
+  namespace: ${local.cert_manager_namespace}
+type: Opaque
+stringData:
+# TODO: The cf_api_token variable is also used by the Cloudflare Terraform Proider, we should use a different here.
```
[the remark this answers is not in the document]
MaxSchaefer: No reason. Fixed.

kubernetes/iac/environments/staging/cert.tf
MaxSchaefer: You mean `sso.helpwave.de` and `staging-sso.helpwave.de`? They are proxied via cloudflare.

FoseFx (approved):

kubernetes/iac/modules/hw-tf-module-cluster-resources/apisix.tf
MaxSchaefer: Me neither. The goal of this pull request is to be on the same level as with the previous HTTP approach. In the future we should not expose the "Dapr way" of gRPC, thats right. If mobile-app can adopt gRPC-web then we should not expose both, gRPC-web and gRPC.

MaxSchaefer:
> Sorry for the delay! I did not quite get the benefits of switching to the kubectl provider. Can you link the medium article you mentioned?
> 
> Can you also compile a list of all TODOs that are left after after this PR? It's important to not lose track of them

What delay? 😄 Don't worry. I've added the [Medium article](https://medium.com/@danieljimgarcia/dont-use-the-terraform-kubernetes-manifest-resource-6c7ff4fe629a).
If we apply terraform to an "empty world", there is no cluster (therefore no ClusterAPI). The cluster itself and the resources inside of the cluster should be created by terraform. But in the plan phase of terraform, before any api requests to create resources, the kubeneretes provider wants a ClusterAPI to plan out the resource changes inside the cluster, but there is no cluster yet.

Another solution is to split out the cluster-resources to another terraform configuration. But that looks like too much overhead.

MaxSchaefer:
To-dos:
- [Why bypassing Cloudflare and using cert-manager?](https://github.com/helpwave/services/pull/74/files#diff-6c9ab76d80d3b257c2398d71343f541fe23100f61db9c56324bf5a39ce4f13d3R24)
- [Also match against hostnames](https://github.com/helpwave/services/pull/74/files#diff-abec648cc46294ab73d4030913cf9356a291f5eaf8b3d6c2e7f348a9c04f2562R133)

kubernetes/iac/environments/staging/cert.tf
FoseFx: Yes, I forgot I renamed it 😅. Keycloak itself requires a certificate (and won't start without it). There are ways around it, but they end up requiring a lot more configuration (i.e. disabling issuer verification and implementing it manually), including on every service, which I try to minimize. So when we have a cert for those it would simplify things. I though I opened an issue for this, but apparently I did not.

PR #75, opened 2023-02-19
feat(task-svc): bootstrap task service

### LINKED ISSUE #64: Bootstrap Tasks Service

MaxSchaefer:
This pull-request introduces the `task-svc`.
In a meeting, we discussed that we should not slice this service as much. In its MVP form, this service should include `tasks`, `patients`, `beds`, `wards`. With this pull-requests, the service gets bootstrapped with `tasks` and a basic `patient` implementation.

- The different gRPC services are split into `internal/` packages.
- I don't really like the `uuid.NullUUID` handling in `internal/task`.
- I have intentionally not introduced database relations to simplify future slicing. We then only need to replace the database calls with RPC calls.
- To-dos
  - Auth
  - Check if user exists in `AssignTaskToUser`

---
Closes #64


services/task-svc/migrations/000001_init.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE TABLE IF NOT EXISTS tasks (
+	id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	name text NOT NULL,
+	description text  NOT NULL DEFAULT '',
+    status integer NOT NULL,
```
FoseFx, line 7: Formatting, also the brackets

Also an integer is not a great way to store this. We should either use an enum, a string or separate tables.

services/task-svc/migrations/000001_init.down.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS tasks;
+DROP TABLE IF EXISTS patients;
```
FoseFx, line 2: Not a problem here, but in general it is better to uphold the convention of dropping things in the reverse order they were created, to prevent constraint issues

services/task-svc/main.go
```diff
@@ -0,0 +1,40 @@
...
+	)
+
+	port := hwutil.GetEnvOr("PORT", "8080")
+	addr := hwutil.GetEnvOr("ADDR", fmt.Sprintf(":%s", port))
+
+	common.StartNewGRPCServer(addr, func(server *grpc.Server) {
+		api.RegisterTaskServiceServer(server, task.NewServiceServer())
+		api.RegisterPatientServiceServer(server, patient.NewServiceServer())
+		api.RegisterWardServiceServer(server, ward.NewServiceServer())
+		reflection.Register(server)
```
FoseFx, line 38: maybe move this into `StartNewGRPCServer`, so reflection is available for all services, what would the implications be?

services/task-svc/internal/patient/patient.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/google/uuid"
+	zlog "github.com/rs/zerolog/log"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"hwgorm"
+	"task-svc/api"
+)
+
+type Base struct {
+	HumanReadableIdentifier string `gorm:"column:human_readable_identifier"`
```
FoseFx, line 14: is `gorm:column` really needed here (and below)?

services/task-svc/api/patient-svc.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+option go_package = "task-svc/api";
+
+//
+// CreatePatient
+//
+
+message CreatePatientRequest {
+	string humanReadableIdentifier = 1;
```
FoseFx, line 10: don't we need any kind of validation here? I think this will accept an empty string or, by extend a missing value. please check for all other requests as well. [see this example](https://github.com/helpwave/services/blob/main/services/auth-svc/api/auth-svc.proto#L21)

services/task-svc/api/patient-svc.proto
```diff
@@ -0,0 +1,49 @@
...
+message CreatePatientResponse {
+	string id = 1;
+}
+
+//
+// GetPatient
+//
+
+message GetPatientRequest {
+	string id = 1;
```
FoseFx, line 22: you can validate id is a uuid here already

services/task-svc/internal/patient/patient.go
```diff
@@ -0,0 +1,101 @@
...
+	// TODO: Auth
+
+	id, err := uuid.Parse(req.Id)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	patient := Patient{Id: id}
+	result := db.First(&patient)
+	if err := result.Error; err != nil {
```
FoseFx, line 70: I don't know how this will behave when the query returns empty. This will result in a bad UX, when we respond with an internal server error. Can we handle this case?

I think in hwgorm there is a helper function to separate user and client errors, but it maybe was lost in hw/rest-api

services/task-svc/internal/task/task.go
```diff
@@ -0,0 +1,240 @@
...
+	// TODO: Auth
+
+	id, err := uuid.Parse(req.Id)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	task := Task{Id: id}
+	result := db.First(&task)
+	if err := result.Error; err != nil {
```
FoseFx, line 88: see my comment on getpatient

services/task-svc/api/task-svc.proto
```diff
@@ -0,0 +1,107 @@
...
+
+//
+// TaskService
+//
+
+service TaskService {
+	rpc CreateTask (CreateTaskRequest) returns (CreateTaskResponse) {}
+	rpc GetTask (GetTaskRequest) returns (GetTaskResponse) {}
+	rpc UpdateTask (UpdateTaskRequest) returns (UpdateTaskResponse) {}
+	rpc TaskToInProgress (TaskToInProgressRequest) returns (TaskToInProgressResponse) {}
```
FoseFx, line 103: A single `ChangeStatus` request may be the simpler solution here. We could also simply move this into the Update endpoint

services/task-svc/internal/task/task.go
```diff
@@ -0,0 +1,240 @@
...
+	// TODO: Auth
+
+	patientId, err := uuid.Parse(req.PatientId)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// Check if patient exists
+	if err := db.First(intPatient.Patient{Id: patientId}).Error; err != nil {
+		return nil, status.Error(codes.NotFound, err.Error())
```
FoseFx, line 48: the error here may be arbitrary (i.e. connectivity issues), but we always return a NotFound. This will make our lives harder when monitoring. (Happens a bunch of times here)

services/task-svc/internal/task/task.go
```diff
@@ -0,0 +1,240 @@
...
+
+	task := Task{Id: id}
+	result := db.First(&task)
+	if err := result.Error; err != nil {
+		log.Warn().Err(err).Msg("database error")
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	assignedUserId := task.AssignedUserId.UUID.String()
+	if !task.AssignedUserId.Valid {
```
FoseFx, line 94: is this needed? wont it be a zero-type else and NULLed by gorm

services/task-svc/internal/task/task.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// TODO: Auth
+
+	id, err := uuid.Parse(req.Id)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	task := Task{Id: id}
+	updates := Task{Base: Base{Status: 1}}
```
FoseFx, line 142: dont use the constant value (1) directly here, use the enum instead

FoseFx (changes requested):

services/task-svc/api/task-svc.proto
MaxSchaefer: I wanted to be very verbose about the users intention of the status change. In that case, the endpoint can also just dispatch an event to other services.

services/task-svc/internal/patient/patient.go
MaxSchaefer: We should be explicit.

services/task-svc/main.go
MaxSchaefer: `StartNewGRPCServer` enables gRPC reflection during development mode.

services/task-svc/internal/task/task.go
MaxSchaefer: Yes, right. I've adopted the `hwgorm.IsOurFault` helper. Thanks for the suggestion.

services/task-svc/internal/task/task.go
MaxSchaefer: It's the `uuid.NullUUID` handling. How should we handle this?

services/task-svc/internal/task/task.go
MaxSchaefer: We should commit to an error pattern and apply that across the services. Right now, it's not that consistent.

services/task-svc/internal/task/task.go
FoseFx: Oh I though this would prepare a db query 😅

But now I think there is a different issue here:
If it can be null we should mark it as `optional` in the .proto, I think the go generator will then let us work with pointers, and we should use nil in this case

libs/common/grpc.go
```diff
@@ -47,6 +48,11 @@ func StartNewGRPCServer(addr string, registerServerHook func(*grpc.Server)) {
 
 	registerServerHook(server)
 
+	if hwutil.MatchEnv("MODE", DevelopmentMode) {
```
FoseFx, line 51: you should use [`common.Mode`](https://github.com/helpwave/services/blob/main/libs/common/setup.go#L24) instead

services/task-svc/internal/task/task.go
MaxSchaefer: In postgres a null uuid gets represented by `00000000-0000-0000-0000-000000000000`. The stringification of `uuid.NullUUID` should handle the translation of `00000000-0000-0000-0000-000000000000` to `""` but this seems not to work. I've added a code comment.

services/task-svc/migrations/000001_init.up.sql
FoseFx: there is some kind of formatting error here

```suggestion
       status integer NOT NULL,
```

services/task-svc/main.go
```diff
@@ -0,0 +1,38 @@
...
+	"hwgorm"
+	"hwutil"
+	"task-svc/api"
+	"task-svc/internal/patient"
+	"task-svc/internal/task"
+	"task-svc/internal/ward"
+)
+
+const ServiceName = "task-svc"
+
```
FoseFx, line 16:
```suggestion

// Version is set at compile time
```

services/task-svc/internal/patient/patient.go
FoseFx: [GORM Docs:](https://gorm.io/docs/models.html#Conventions)
> GORM prefers convention over configuration. [...] If you follow the conventions adopted by GORM, you’ll need to write very little configuration/code.

I personally don't really care if we are explicit or not, as long as we are consistent. So we should either add the gorm tags to all existing models or remove them here.

services/task-svc/internal/patient/patient.go
```diff
@@ -0,0 +1,103 @@
...
+	"task-svc/api"
+)
+
+type Base struct {
+	HumanReadableIdentifier string `gorm:"column:human_readable_identifier"`
+}
+
+type Patient struct {
+	Base
+	Id uuid.UUID `gorm:"column:id"`
```
FoseFx, line 19: `ID` is used in other models (uppercase d)

services/task-svc/internal/patient/patient.go
```diff
@@ -0,0 +1,103 @@
...
+
+	id, err := uuid.Parse(req.Id)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	patient := Patient{Id: id}
+	if err := db.First(&patient).Error; err != nil {
+		if hwgorm.IsOurFault(err) {
+			log.Warn().Err(err).Msg("database error")
```
FoseFx, line 70: This is definitely worth an `Error` log level. But keep in mind, that returned errors are already logged by the [logging middleware](https://github.com/helpwave/services/blob/main/libs/common/grpc.go#L127). I'll open an issue to change the log level for interal errors from warn to error.

services/task-svc/internal/task/task.go
```diff
@@ -0,0 +1,247 @@
...
+	assignedUserId := task.AssignedUserId.UUID.String()
+	if !task.AssignedUserId.Valid {
+		assignedUserId = ""
+	}
+
+	return &api.GetTaskResponse{
+		Id:             task.Id.String(),
+		Name:           task.Name,
+		Description:    task.Description,
+		Status:         api.TaskStatus(task.Status),
```
FoseFx, line 109: "Redundant type conversion"

services/task-svc/internal/patient/patient.go
MaxSchaefer: Sounds good! I've removed the log there, didn't know that the logging middleware already does this.

FoseFx (approved):
🥳

PR #80, opened 2023-02-25
chore(common): use log level Error instead of Warn for logging middleware

### LINKED ISSUE #76: Elevate logging level for `codes.Internal` errors in logging middleware
[Currently](https://github.com/helpwave/services/blob/main/libs/common/grpc.go#L127) the logging middleware always logs errors at the `Warn` level.

Some errors, notably those that are `Internal`, should be logged on the `Error` level.

MaxSchaefer:
Closes #76

### DIFF
--- a/libs/common/grpc.go
+++ b/libs/common/grpc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
+	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/reflection"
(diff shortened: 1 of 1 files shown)


libs/common/grpc.go
```diff
@@ -136,7 +136,7 @@ func loggingUnaryInterceptor(ctx context.Context, req interface{}, info *grpc.Un
 		message := statusError.Message()
 		details := statusError.Details()
 		log.
-			Warn().
+			Error().
```
FoseFx, line 139: this is not the expected behavior
> *Some* errors, notably those that are Internal, should be logged on the Error level.

Users may enter invalid input, or all kind of other issues may arise, which are not worthy being an Error.
But if the route handler deems the issue to be caused by an issue within the service (i.e. returning an InternalServerError), it should be an Error, not just a warning.

FoseFx (changes requested):

libs/common/grpc.go
MaxSchaefer: Okay, get it. I've updated the code to resolve the error level by specific errors.

libs/common/grpc.go
```diff
@@ -152,6 +154,16 @@ func loggingUnaryInterceptor(ctx context.Context, req interface{}, info *grpc.Un
 	return res, err
 }
 
+// Primary used to raise the log level on specific errors
```
FoseFx, line 157:
```suggestion
// resolveLogLevelForError is primary used to raise the log level on specific errors
```

FoseFx (approved):

PR #84, opened 2023-02-28
feat(user-svc): add personal organization after user creation

### LINKED ISSUE #51: Create personal organization
https://github.com/helpwave/helpwave/issues/13:

This is blocked by #50. Simply create a new organization with `is_personal` set to true on behalf of a newly created user, after creating said user.

FoseFx:
Closes #51

### DIFF
--- a/services/user-svc/main.go
+++ b/services/user-svc/main.go
@@ -113,6 +113,19 @@ func (userServiceServer) CreateUser(ctx context.Context, request *api.CreateUser
 		log.Info().Str("userID", userID).Msg("created new user")
 	}
 
+	// Create personal organization
+	_, err = createOrganization(ctx, &api.CreateOrgRequest{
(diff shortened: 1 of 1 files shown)


MaxSchaefer (changes requested):
With this approach, when the later operation (`createOrganization()`) fails, the creation of the user gets not rolled back or the creation of the organization is not tried again.

To combine these two atomic operations, the `CreateUser()` endpoint should raise an `UserCreated` event. One of the handlers for that event should tries to call `createOrganization()`.

Lets use [Dapr Pub/Sub](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-overview/) for that.

MaxSchaefer:
@FoseFx you closed this. What are your next steps?

FoseFx:
Re-implement it using pub/sub as you suggested

PR #85, opened 2023-03-03
issue/51 personal organizations

### LINKED ISSUE #51: Create personal organization

FoseFx:
Closes #51


services/user-svc/main.go
```diff
@@ -114,6 +123,15 @@ func (userServiceServer) CreateUser(ctx context.Context, request *api.CreateUser
 		log.Info().Str("userID", userID).Msg("created new user")
 	}
 
+	// prepare for event
+	user.ID = &userID
+	user.Credentials = nil
+
+	// publish USER_CREATED event
+	if err := daprClient.PublishEvent(ctx, "pubsub", "USER_CREATED", user); err != nil {
+		log.Error().Err(err).Msg("could not publish USER_CREATED event")
```
MaxSchaefer, line 132: It is less likely that publishing the event will fail instead of the keycloak operation. But we have the same issue as before. When the creation of the event fails, the user was created, the organization gets not created. Okay for now.
`TODO: I guess the Transactional Outbox Pattern is a good use for that.`

MaxSchaefer (changes requested):

services/user-svc/main.go
MaxSchaefer: As I said, okay for now. I've added the comment.

services/user-svc/api/user-events.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+option go_package = "user-svc/api";
+
+// TOPIC: USER_CREATED
+message CreateUserEvent {
```
MaxSchaefer, line 6:
```suggestion
message UserCreatedEvent {
```

MaxSchaefer (changes requested):

services/user-svc/api/user-events.proto
MaxSchaefer: When this events gets dispatched, the user was already created.

MaxSchaefer (approved):

PR #90, opened 2023-03-22
Update User

### LINKED ISSUE #49: Basic User Management Endpoints
- [ ] update full and nick name
- [ ] change E-Mail address
- [ ] change password

FoseFx:
Closes #49

Adds a single update route which allows the client to change:
- fullname
- nickname
- email (as username = email, also the username)
- password


services/user-svc/api/user-svc.proto
```diff
@@ -18,6 +18,21 @@ message CreateUserResponse {
...
+//
+// Update User
+//
+
+message UpdateUserRequest {
+	// if userId is missing the requesting user's id is used
+	optional string userId = 1; // @gotags: validate:"omitempty,uuid4"
+	optional string fullName = 2;
+	optional string nickname = 3;
+	optional string email = 4; // @gotags: validate:"omitempty,email"
```
MaxSchaefer, line 30: Can you remove email here? Because of the side effects of an email change, we should handle this in a separate RPC.

services/user-svc/api/user-svc.proto
```diff
@@ -18,6 +18,21 @@ message CreateUserResponse {
 	string userID = 1;
 }
 
+//
+// Update User
+//
+
+message UpdateUserRequest {
+	// if userId is missing the requesting user's id is used
+	optional string userId = 1; // @gotags: validate:"omitempty,uuid4"
```
MaxSchaefer, line 27: This should be required.

MaxSchaefer (changes requested):

services/user-svc/api/user-svc.proto
FoseFx: Why?

MaxSchaefer: Sorry, my bad. Didn't catch the `hwutil.DerefStringOr`.

services/user-svc/api/user-svc.proto
MaxSchaefer: We should try to intercept the user's intent already at the API level. Also, in the future, we will publish an event like `EmailUpdated`. In such endpoints, we would then have quite a few if-else conditions that publish events. Since it is currently not particularly necessary, feel free to resolve this conversation, but we will revisit this endpoint in the future.

MaxSchaefer (approved):

PR #98, opened 2023-03-30
refactor protobufs

### LINKED ISSUE #87: Refactor .proto files
Now proto files are bound to a service by our toolchain. This may cause problems, for example event protos don't semantically belong to a service.

Create a `/proto` dir and move protos there.
The fs may look like this:

```
lib/
 events/
   user-events.pb.go
services/
  user-svc/
    api/
      user-svc.pb.go
      custom.go
proto/
  user-svc/
    user-svc.proto
  events/
    user-events.proto
  utility-messages.proto
```

Also consider [buf.build](buf.build)

FoseFx:
Closes #87

**Changes**
- [x] use buf
- [x] make go packages work
- [x] add gotags plugin
- [x] replace with buf docker image
- [x] set up CI for buf lint and build
- [x] merge main

Future work:
- [ ] fix buf lints
- [ ] `buf breaking` ci


MaxSchaefer:
Is `buf` a drop-in replacement for `protoc`? Does the switch to generated protos by `buf` introduces breaking API changes?

FoseFx:
> Is `buf` a drop-in replacement for `protoc`?

Yes

> Does the switch to generated protos by `buf` introduces breaking API changes?

[No](https://buf.build/docs/reference/internal-compiler/)

.github/workflows/buf.yaml
```diff
@@ -0,0 +1,34 @@
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - run: make proto
+      - uses: EndBug/add-and-commit@v9
+        id: commit
+        with:
+          add: 'gen'
+          default_author: github_actions
+          message: 'fix: generate protobufs'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
MaxSchaefer, line 34: If the output of `make proto` is different from what was committed, the ci must fail. When this is the case, a failed `generate-protos / validate-protos` ci indicated that the application was likely developed with a different version of the protos.

Makefile
```diff
@@ -7,14 +7,22 @@ GO_SERVICES = $(subst services/,,$(wildcard services/*))
...
 DOCKER_IMAGES = $(subst images/,,$(wildcard images/*))
 
+.PHONY: proto
+proto:
+	docker run -v $$(pwd):/wd ghcr.io/helpwave/service-preproc:edge lint || true
+	docker run -v $$(pwd):/wd ghcr.io/helpwave/service-preproc:edge generate
+
+.PHONY: proto_lint
+proto_lint:
+	docker run -v $$(pwd):/wd ghcr.io/helpwave/service-preproc:edge lint
```
MaxSchaefer, line 17: Please add an `--rm` to remove the container after the execution.

services/task-svc/internal/room/room.go
```diff
@@ -116,5 +116,13 @@ func (ServiceServer) AddBedsToRoom(ctx context.Context, req *api.AddBedsToRoomRe
 		Int("beds", len(beds)).
 		Msg("beds added")
 
-	return &api.AddBedsToRoomResponse{}, err
+	return &pb.AddBedsToRoomResponse{}, err
+}
+
+func BedsToBedsOfRoom(beds []models.Bed) []*pb.BedOfRoom {
```
MaxSchaefer, line 122: In my last pull request, this was moved to `api/`. Any reason why this is added again to `room/`?

MaxSchaefer (changes requested):
Nice work 🎉
But we should reconsider these non-proto `helpers.go` files.
After `rm -rf gen && make proto` I've got the following changes; I've expected none.
```bash
deleted:    gen/README.md
deleted:    gen/go.mod
deleted:    gen/go.sum
deleted:    gen/proto/services/auth-svc/logging.go
deleted:    gen/proto/services/emergency-room-svc/helpers.go
deleted:    gen/proto/services/task-svc/helpers.go
deleted:    gen/proto/services/user-svc/logging.go
```

services/task-svc/internal/room/room.go
FoseFx: I believe we can't move it to `gen`, as `task-svc` depends on `gen` and this would introduce an import of `task-svc` in `gen`. Cyclic Dependency Graphs don't compile in go AFAIK.

FoseFx:
> Nice work tada But we should reconsider these non-proto `helpers.go` files. After `rm -rf gen && make proto` I've got the following changes; I've expected none.
> 
> ```shell
> deleted:    gen/README.md
> deleted:    gen/go.mod
> deleted:    gen/go.sum
> deleted:    gen/proto/services/auth-svc/logging.go
> deleted:    gen/proto/services/emergency-room-svc/helpers.go
> deleted:    gen/proto/services/task-svc/helpers.go
> deleted:    gen/proto/services/user-svc/logging.go
> ```

I don't like it too much either, but there is no need to remove the whole gen directory. `make proto` will overwrite existing files. This may only be a problem when proto files are renamed / the file structure changes. In that case we can remove all generated go files using `rm gen/**/*.pb.go`.

MaxSchaefer:
> I don't like it too much either, but there is no need to remove the whole gen directory. `make proto` will overwrite existing files. This may only be a problem when proto files are renamed / the file structure changes. In that case we can remove all generated go files using `rm gen/**/*.pb.go` before.

Okay, sounds reasonable for now. This is a to-do for the future.
@FoseFx you want to fix the ci in a later pull request?

FoseFx:
> you want to fix the ci in a later pull request?

The CI works fine, we just violate a lot of linting rules. I want to fix those in a follow-up, yes.

MaxSchaefer (changes requested):
... upsi

MaxSchaefer (approved):
🚀🌊

PR #111, opened 2023-04-07
fix: use correct protos

MaxSchaefer:
`emergency-room-svc` and `user-svc` failed to build

### DIFF
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -24,17 +24,9 @@ jobs:
       deployments: write
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
       - run: make proto
-      - name: Get changed files
-        id: changed-files
-        uses: tj-actions/changed-files@v35
-        with:
-          files: gen
-      - name: Break if files have changed
-        if: steps.changed-files-specific.outputs.any_changed == 'true'
-        run: exit 1
+      - name: Break if generated protos are not checked in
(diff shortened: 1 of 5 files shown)


FoseFx (approved):

MaxSchaefer:
The ci should fail. On local, I have the following changes:
```
modified:   gen/proto/services/auth_svc/v1/
modified:   gen/proto/services/auth_svc/v1/auth_svc_grpc.pb.go
```

.github/workflows/buf.yaml
```diff
@@ -24,17 +24,9 @@ jobs:
...
       - run: make proto
-      - name: Get changed files
-        id: changed-files
-        uses: tj-actions/changed-files@v35
-        with:
-          files: gen
-      - name: Break if files have changed
-        if: steps.changed-files-specific.outputs.any_changed == 'true'
-        run: exit 1
+      - name: Break if generated protos are not checked in
```
FoseFx, line 28:
```suggestion
      - name: Break if generated go modules are not checked in
```

.github/workflows/buf.yaml
```diff
@@ -24,17 +24,9 @@ jobs:
...
+        run: git diff --quiet gen
```
FoseFx, line 29:
```suggestion
        # exits with 1 if there were differences, and 0 if there were none
        run: git diff --quiet gen
```

FoseFx (approved):

PR #121, opened 2023-04-08
fix: add package-json, was removed in migration

MaxSchaefer:


gen/ts/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  	"name": "@helpwave/proto-ts",
```
use-to, line 2:
```suggestion
 	"name": "@helpwave/proto-ts",
```

use-to (changes requested):

use-to (approved):

PR #130, opened 2023-04-13
feat: introduce ory-svc

MaxSchaefer:
This pull request introduces the http microservice `ory-svc`. This services should handle Ory related stuff like inbound http webhooks. Since I first wanted to make a service-to-service call from `ory-svc` to `user-svc` to create an organization, this pr also contains an example for service-to-service communication via Dapr and gRPC. Now, however, the `ory-svc` raises the `UserCreatedEvent` after a successful registration, whereupon the `user-svc` creates the personal organization for this user.


libs/hwutil/util.go
```diff
@@ -0,0 +1,9 @@
+package hwutil
+
+import "fmt"
+
+func GetAddr() string {
```
FoseFx, line 5: bad function name. Also, it may be better to make it part of `common`

proto/libs/events/v1/organization_events.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package proto.libs.events.v1;
+
+option go_package = "gen/proto/libs/events";
+
+// TOPIC: ORGANIZATION_CREATED
+message OrganizationCreatedEvent {
+	string id = 1;
```
FoseFx, line 9:
```suggestion
	// the new organization's id
	string id = 1;
```

services/ory-svc/.env.example
```diff
@@ -0,0 +1,27 @@
...
+LOG_LEVEL=debug
+
+# The port on which the the grpc service will listen on (default: 8080)
+PORT=8080
+
+# port on which the dapr sidecar listens on
+# in prod this is managed by the dapr-sidecar-injector
+DAPR_HTTP_PORT=
+
+#
```
FoseFx, line 17: we don't need this anymore, I assume

services/ory-svc/README.md
```diff
@@ -0,0 +1,3 @@
+# ory-svc
+
+> Service to handle Ory related in-/outbound operations
```
FoseFx, line 3:
```suggestion
> Service to handle [Ory](https://ory.sh/) related in-/outbound operations
```

services/ory-svc/main.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if err := service.Start(); err != nil {
+		zlog.Fatal().Str("addr", addr).Err(err).Msg("could not start http server")
+	}
+}
+
+func afterRegistrationWebhookHandler(ctx context.Context, in *daprcmn.InvocationEvent) (out *daprcmn.Content, err error) {
+	if in.Verb != http.MethodPost {
+		return nil, status.Error(codes.InvalidArgument, "invalid method")
```
FoseFx, line 49: if an err is returned, dapr will always assume something went wrong server-side and return a 500 [(relevant code)](https://github.com/dapr/go-sdk/blob/c5377994aa5f6ffd6d64fc3ee7d2bba7a9124018/service/http/invoke.go#L67-L69).

other than that, we return a status.Error here, which is from gRPC land, though not really relevant 
it is semantically wrong and may give the wrong impression when reading the code

last thing: because this is not part of the grpc chain I don't think it is using the logger middleware to log returned errors (though I might be wrong here)

services/ory-svc/main.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if in.ContentType != "application/json" {
+		return nil, status.Error(codes.InvalidArgument, "invalid content-type")
+	}
+
+	var payload pb.AfterRegistrationWebhookPayload
+	if err := json.Unmarshal(in.Data, &payload); err != nil {
```
FoseFx, line 57: use [`hwutil.ParseValidJson(...)`](https://github.com/helpwave/services/blob/38021b3da9f7be051bc3c8abc5ec56cb69c94077/libs/hwutil/parse.go#L13) instead to validate the payload as well

services/user-svc/README.md
```diff
@@ -1,4 +1,3 @@
 # user-svc
 
 > User management service
-
```
FoseFx, line 4: I am a big fan of trailing new lines, looks like your IDE isn't ^^

proto/services/user_svc/v1/organization_svc.proto
```diff
@@ -12,12 +12,29 @@ message CreateOrganizationRequest {
...
 
 message CreateOrganizationResponse {
 	string id = 1;
 }
 
+//
+// CreateOrganizationForUser
+//
+
+message CreateOrganizationForUserRequest {
```
FoseFx, line 26: why not extend `CreateOrganizationRequest` with an optional `user_id` field?

services/user-svc/README.md
MaxSchaefer: Mh, weird. My IDE + editorconfig doesn't complain, and they show me a newline on my end ... when you pull?

services/auth-svc/.env.example
```diff
@@ -13,10 +13,6 @@ PORT=8080
 # The port on which the http server will listen on (default: 8081)
 HTTP_PORT=8081
 
-# port on which the dapr sidecar listens on
-# in prod this is managed by the dapr-sidecar-injector
-DAPR_HTTP_PORT=
-
 #
 # Keycloak / OIDC config
```
FoseFx, line 17: I think github chose to only show the code above the line I commented on. I meant everything below it though. The HTTP_PORT can stay

services/ory-svc/README.md
```diff
@@ -0,0 +1,29 @@
...
+To [customize the request body of the webhook](https://www.ory.sh/docs/guides/integrate-with-ory-cloud-through-webhooks#customizing-request-body-with-jsonnet) that comes from Ory we need to use Jsonnet.
+This file gets Base64 encoded and used as a string in the `identity-config`.
+The results look like this:
+```yaml
+...
+oidc:
+  hooks:
+  - hook: session
+  - hook: web_hook # helpwave/services/ory-svc
+	config:
```
FoseFx, line 24: Are these tabs?

services/ory-svc/main.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, newErrAndLog(ctx, "invalid method")
+	}
+
+	if in.ContentType != "application/json" {
+		return nil, newErrAndLog(ctx, "invalid content-type")
+	}
+
+	var payload pb.AfterRegistrationWebhookPayload
+	if err := hwutil.ParseValidJson(in.Data, &payload); err != nil {
+		return nil, err
```
FoseFx, line 63:
```suggestion
		return nil, newErrAndLog(ctx, err.Error())
```

services/ory-svc/main.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	userID, err := uuid.Parse(payload.UserId)
+	if err != nil {
+		return nil, err
```
FoseFx, line 68:
```suggestion
		return nil, newErrAndLog(ctx, err.Error())
```

services/ory-svc/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+	userCreatedEvent := &events.UserCreatedEvent{
+		Id:       userID.String(),
+		Email:    payload.Email,
+		Nickname: payload.Nickname,
+		Name:     payload.Name,
+	}
+
+	if err := common.PublishMessage(ctx, daprClient, DaprPubsub, "USER_CREATED", userCreatedEvent); err != nil {
+		return nil, err
```
FoseFx, line 79:
```suggestion
		return nil, newErrAndLog(ctx, err.Error())
```

FoseFx (approved):

services/auth-svc/.env.example
MaxSchaefer: We will clean this up when we are about to remove Keycloak entirely.

PR #139, opened 2023-04-15
feat: replace Keycloak with generic OIDC provider

### LINKED ISSUE #107: Adopt generic OIDC for authentication
Ory acts as a generic OAuth2.0 / OIDC provider. Our services should be able to verify the validity of the OIDC JWTs. Our authentication middleware should also be able to validate tokens by a predefined shared secret __only in development mode__.

MaxSchaefer:
This pull request shifts the entire authentication to our OIDC provider by removing Keycloak and becoming more OIDC provider-agnostic. Most of the hard-coded default values (configurable via envs) are for Ory, our provider of choice.

**Fake tokens**
A fake token is a Base64 encoded json structure in the schema of `IDTokenClaims`. This is not our discussed approach [ADR-008: Fake OIDC ID-Token](https://wiki.helpwave.de/doc/adr-008-fake-oidc-id-token-ITuLg5F8bC), but I find it quite practical. When we're accepting fake tokens, we don't really need to verify the signature. With this approach, the clients are more flexible during testing because they can define their own values for `IDTokenClaims` without going through an endpoint of `services`. _I wil update ADR-008 in the near future_

Closes #107


libs/common/auth.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return nil
+}
+
+func isAuthSetUp() bool {
+	return oauthConfig != nil && verifier != nil
+}
+
+func setupAuth() {
+	issuerUrl := hwutil.GetEnvOr("OAUTH_ISSUER_URL", DEFAULT_OAUTH_ISSUER_URL)
```
FoseFx, line 62: can you add it to the .env.example files for documentation?

libs/common/auth.go
```diff
@@ -0,0 +1,113 @@
...
+	if issuerUrl != DEFAULT_OAUTH_ISSUER_URL {
+		zlog.Warn().Str("OAUTH_ISSUER_URL", issuerUrl).Msg("Using auth without the default OAuth issuer url")
```
FoseFx, line 64:
```suggestion
		zlog.Warn().Str("OAUTH_ISSUER_URL", issuerUrl).Msg("using custom OAuth issuer url")
```

libs/common/auth.go
```diff
@@ -0,0 +1,113 @@
...
+
+	verifier = provider.Verifier(&oidc.Config{ClientID: oauthConfig.ClientID})
+}
+
+// VerifyIDToken verifies the correctness of the accessToken and returns its claim.
+// The claim is checked to be as expected.
+// Service must be set up with auth!
+func VerifyIDToken(token string) (*IDTokenClaims, error) {
+	// technically this method is meant to be used to verify identity tokens,
+	// but we use it for access tokens instead as keycloak uses similar JWTs for access tokens
```
FoseFx, line 88: mentions keycloak, does this still hold for ory?

libs/common/setup.go
```diff
@@ -49,7 +51,12 @@ func Setup(serviceName, version string, auth bool) {
 	}
 
 	if auth {
-		setupKeycloak()
+		if strings.ToLower(hwutil.GetEnvOr("FAKE_TOKEN_ENABLE", "false")) == "true" && Mode == DevelopmentMode {
```
FoseFx, line 54: staging will run in production mode, yet it should accept fake tokens, right?

Also maybe prefix the env name with INSECURE_

libs/common/setup.go
```diff
@@ -49,7 +51,12 @@ func Setup(serviceName, version string, auth bool) {
...
+			FakeTokenEnable = true
+			log.Warn().Msg("FAKE_TOKEN_ENABLE is set to true, accepting fake tokens")
```
FoseFx, line 56: I think this should be an error-level log. If we see this in prod, it's an error and noticed faster. In dev environments, the devs know it can be ignored.

libs/common/grpc.go
```diff
@@ -74,34 +77,85 @@ func authFunc(ctx context.Context) (context.Context, error) {
...
+		return nil, status.Errorf(codes.Unauthenticated, "no auth token: %v", err)
 	}
 
 	// verify token
-	claims, err := VerifyAccessToken(token)
+	claims, err := VerifyIDToken(token)
+
+	// If FakeTokenEnable is true and Mode is development,
+	// we accept unverified Base64 encoded json structure in the schema of IDTokenClaims as well.
+	// This allows the client to pass self-defined "Fake ID Token Claims" without go through our auth provider.
```
FoseFx, line 88:
```suggestion
	// This allows the client to pass self-defined "Fake ID Token Claims" without going through our auth provider.
```

FoseFx (approved):

MaxSchaefer:
Thanks @FoseFx for the review. Good idea to raise the log level and prefixing the env with `INSECURE_`. Just to be sure, I've re-requested the review.

FoseFx (approved):

PR #136, opened 2023-04-14
Adapt Protobuf Definitions

use-to:
Closes #131
Closes #133 
Closes #134

PR #140, opened 2023-04-18
remove keycloak

FoseFx:
Blocked by #139

I kept the cert and coredns deployments for future use


MaxSchaefer:
@FoseFx thanks! Nice clean-up 🧹 I guess it makes sense to [add the required envs](https://github.com/helpwave/services/blob/main/libs/common/auth.go#LL61C4-L61C4) for auth related stuff as well.

FoseFx:
I think the lint errors are due to #136?

MaxSchaefer:
> I think the lint errors are due to #136?

Yes. Breaking changes in the defition. It looks like that these breaking changes occured int the `room-svc`. You can either fix these issues in this pr or wait for #147 with this pr in draft. I'll prefer the first option.

use-to:
Our action detecting changed files is not working correctly.
This error should have occurred right after we updated protobuf definitions.

MaxSchaefer (approved):

PR #142, opened 2023-04-18
fix(ci): add `check-preconditions` dependency back

FoseFx:
Likely fixes the root cause for this run failing: https://github.com/helpwave/services/actions/runs/4709945814/jobs/8353316735


MaxSchaefer (approved):
Mhnnn 😅 Thanks!

PR #177, opened 2023-06-04
fix: rename to UnPublish

### LINKED ISSUE #176: rename Unpublish to UnPublish
Unpublish should be UnPublish instead.

florian-sabonchi:
closes #176


use-to (changes requested):
https://github.com/helpwave/services/issues/176

PR #181, opened 2023-06-08
implement memberships

### LINKED ISSUE #158: Implement `Memberships`
A relational diagram of the membership's entity can be found at https://github.com/helpwave/services/issues/149.

DasProffi:
closes #158


services/user-svc/internal/organization/organization.go
```diff
@@ -175,13 +176,65 @@ func (s ServiceServer) GetOrganization(ctx context.Context, req *pb.GetOrganizat
...
+	}
+
+	if err := AddUserToOrganization(ctx, db, userID, organizationID); err != nil {
+		return nil, err
+	}
+
+	// Make the creating user admin of the organization
+	if err := ChangeMembershipAdminStatus(db, userID, organizationID, false); err != nil {
+		return nil, err
+	}
```
MaxSchaefer, line 199: The default value of `IsAdmin` is already false. I guess we don't really need to do this twice.

MaxSchaefer (changes requested):

services/user-svc/internal/organization/organization.go
```diff
@@ -349,3 +407,16 @@ func HandleUserCreatedEvent(ctx context.Context, evt *daprcmn.TopicEvent) (retry
 
 	return false, nil
 }
+
+func ChangeMembershipAdminStatus(db *gorm.DB, userID uuid.UUID, organizationID uuid.UUID, isAdmin bool) error {
```
MaxSchaefer, line 411: [To reply](https://github.com/helpwave/services/issues/158#issuecomment-1584086207)
The configuration of roles (in that case the `IsAdmin` flag) should not be in concern of `AddUserToOrganization()`, non-admin by default, least privileges.

That `AddUserToOrganization()` is untouched also comes handy for the current implementation of invitations.

DasProffi: The is admin isn't done in there, but the renaming of the struct is necessary, because the table changed

MaxSchaefer (approved):

PR #182, opened 2023-06-08
add is_discharged attribute to patients database and struct

### LINKED ISSUE #163: Discharged patients
Add a boolean called `dischaged` (default `false`) to the patient and set the value to `true` on [DischargePatientRequest](https://proto.helpwave.de/#proto.services.task_svc.v1.DischargePatientRequest).

DasProffi:
closes #163


services/task-svc/migrations/000009_init.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE patients
+	ADD is_discharged INTEGER NOT NULL DEFAULT 0;
```
MaxSchaefer, line 2:
```suggestion
	ADD is_discharged BOOLEAN NOT NULL DEFAULT 0;
```
I mean, `INTEGER` also works ...

MaxSchaefer (approved):

services/task-svc/migrations/000009_init.up.sql
DasProffi: error: migration failed: column "is_discharged" is of type boolean but default expression is of type integer in line 0: ALTER TABLE patients
        ADD is_discharged BOOLEAN NOT NULL DEFAULT 0;
 (details: pq: column "is_discharged" is of type boolean but default expression is of type integer)

PR #198, opened 2023-06-17
Issue/192 decline invitation

### LINKED ISSUE #192: Add `DeclineInvitation` method to organization-svc
This method has the side effect of changing the status of an Invitation, as with issue #180. The state transition should be `Pending` to `Declined`.

florian-sabonchi:
closes #192


services/user-svc/internal/organization/organization.go
```diff
@@ -341,6 +342,77 @@ func (s ServiceServer) AcceptInvitation(ctx context.Context, req *pb.AcceptInvit
...
+			State:          invitation.State,
+		}
+	})
+
+	return &pb.GetInvitationsByUserResponse{
+		Invitations: invitationsResponse,
+	}, nil
+}
+
+func (s ServiceServer) DeclineInvitation(ctx context.Context, req *pb.DeclineInvitationRequest) (*pb.DeclineInvitationResponse, error) {
```
MaxSchaefer, line 384: Check of mail and state transition as in `AcceptInvitation()` is missing. We should wrap the query of an invitation with a function called `GetInvitationByIdAndEmail()`, the state check should be a part of the controller due to the error handling. After that, we could reuse this function for `AcceptInvitation()` as well.

MaxSchaefer (changes requested):

services/user-svc/internal/organization/organization.go
```diff
@@ -487,16 +517,18 @@ func GetOrganizationById(db *gorm.DB, id uuid.UUID) (*Organization, error) {
...
 
-	if err := db.First(&invite).Error; err != nil {
-		return nil, err
+	var invitation Invitation
+	if err := db.Where("id = ? AND email = ?", id, email).First(&invitation).Error; err != nil {
+		if hwgorm.IsOurFault(err) {
+			return nil, status.Error(codes.Internal, err.Error())
+		} else {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invitation with id %s and email %s not found", id, email))
+		}
```
MaxSchaefer, line 528: Just return the gorm error. The gRPC errors should be handled on the method level above.

services/user-svc/internal/organization/organization.go
```diff
@@ -487,16 +517,18 @@ func GetOrganizationById(db *gorm.DB, id uuid.UUID) (*Organization, error) {
 	return &organization, nil
 }
 
-func GetInvitationById(db *gorm.DB, id uuid.UUID) (*Invitation, error) {
```
MaxSchaefer, line 490: Please leave the method as it is. Just add the new one.

MaxSchaefer (changes requested):

services/user-svc/internal/organization/organization.go
florian-sabonchi: But then we have an unused Funktion

MaxSchaefer: Yes. You are right. But this function can be used later.

florian-sabonchi: OK

services/user-svc/internal/organization/organization.go
```diff
@@ -488,15 +518,29 @@ func GetOrganizationById(db *gorm.DB, id uuid.UUID) (*Organization, error) {
...
+
+func GetInvitationByIdAndEmail(db *gorm.DB, email string, id uuid.UUID) (*Invitation, error) {
+
+	var invitation Invitation
+	if err := db.Where("id = ? AND email = ?", id, email).First(&invitation).Error; err != nil {
+		if hwgorm.IsOurFault(err) {
+			return nil, status.Error(codes.Internal, err.Error())
+		} else {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
```
MaxSchaefer, line 540:
```suggestion
		return nil, err
```
Do the error handling inside the "controller", one layer up?

MaxSchaefer (changes requested):

MaxSchaefer (approved):

PR #204, opened 2023-06-24
feat: ward-svc

MaxSchaefer:


services/aggregation-svc/internal/tasks/tasks.go
```diff
@@ -0,0 +1,67 @@
...
+
+		return &pb.GetWardsOverviewResponse_Ward{
+			Id:       ward.Id,
+			Name:     ward.Name,
+			BedCount: bedCount,
+		}
+	})
+
+	return &pb.GetWardsOverviewResponse{Wards: wards}, nil
+}
```
use-to, line 67: I would move that to the ward service. For me, the entire aggregation service is unnecessary.

use-to (changes requested):

services/aggregation-svc/internal/tasks/tasks.go
MaxSchaefer: As already discussed ...
This pr will be migrated to `feat: ward-svc` after the next push

PR #223, opened 2023-06-26
add GetAllTaskTemplates endpoint

### LINKED ISSUE #200: Task templates
This issue tracks the implementation of task templates as an entity in the `task-svc`. In the following, a JSON struct of [a meeting about task templates](https://wiki.helpwave.de/s/2f6e10b9-97c9-407c-aa9c-918915ca1bbb). Right now, task templates does not relate to other tasks.

- When `ward_id` is null, the task template is a personal template.
- When `ward_id` is not null, the task template is a ward template

```json
{
  "id": "string",
  "name": "string",
  "description": "string",
  "ward_id": "string|null", // if set, this task template is on ward level
  "created_by": "string", // mostly for GetAllTaskTemplatesByUser
  "organization_id": "string",
  "subtasks": [
    {
      "id": "string",
      "name": "string",
      "done": "boolean"
    }
  ]
}
```

```[tasklist]
### Tasks
- [ ] #201
- [ ] #208
- [ ] https://github.com/helpwave/services/issues/212
- [ ] https://github.com/helpwave/services/issues/213
- [ ] https://github.com/helpwave/services/issues/214
- [ ] https://github.com/helpwave/services/issues/215
- [ ] #217
- [ ] https://github.com/helpwave/services/issues/216
- [ ] https://github.com/helpwave/services/issues/218
```

### LINKED ISSUE #213: GetAllTaskTemplates (transitive scoped by organization)
This issue should track the implementation of the `GetAllTaskTemplates` method. The method should be scoped by our multi-tenancy policies, e.g. scope by current organization.

DasProffi:
closes #213


services/task-svc/migrations/000011_init.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE task_templates
+	DROP COLUMN organization_id;
```
FoseFx, line 2: IF EXISTS

services/task-svc/migrations/000011_init.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE task_templates
+	ADD organization_id uuid NOT NULL;
```
FoseFx, line 2: if not null, then we should add a DEFAULT for existing rows

proto/services/task_svc/v1/task_template_svc.proto
```diff
@@ -24,6 +24,35 @@ message CreateTaskTemplateResponse {
...
+
+message GetAllTaskTemplatesResponse {
+  repeated TaskTemplate templates = 1;
+
+  message TaskTemplate {
+    string id = 1;
+    string name = 2;
+    string description = 3;
+    bool is_public = 4;
+    optional string user_id = 6;
```
FoseFx, line 43: the name "user_id" is bad. what kind of user? creator / owner? assignee? If I fear this is consistent (in that case it should stay like it is) though

services/task-svc/internal/task-template/task_template.go
```diff
@@ -88,3 +95,44 @@ func (ServiceServer) CreateTaskTemplate(ctx context.Context, req *pb.CreateTaskT
 		Id: taskTemplate.ID.String(),
 	}, nil
 }
+
+func (ServiceServer) GetAllTaskTemplates(ctx context.Context, req *pb.GetAllTaskTemplatesRequest) (*pb.GetAllTaskTemplatesResponse, error) {
+	db := hwgorm.GetDB(ctx)
+
+	organizationId, err := common.GetOrganizationID(ctx)
```
FoseFx, line 102:
```suggestion
	organizationID, err := common.GetOrganizationID(ctx)
```

FoseFx (approved):

services/task-svc/migrations/000011_init.up.sql
DasProffi: what do you suggest as a default, the 0 uuid?

proto/services/task_svc/v1/task_template_svc.proto
DasProffi: we should discuss this whether owner or creator is more appropriate

DasProffi: but since it is used in all other PR's of #200, we should use a seperate issue

services/task-svc/migrations/000011_init.up.sql
FoseFx: If there is no reasonable default, it can not be NOT NULL. We can keep it like it is, if there is no data on staging yet tbh

DasProffi: solved, the 0 uuid can be entered as default with a string

proto/services/task_svc/v1/task_template_svc.proto
DasProffi: see #226

PR #297, opened 2023-07-19
get userID by context for GetOrganizationsByUser

### LINKED ISSUE #296: OrganizationsByUser
### Describe the chore

remove userID parameter from endpoint and instead get it by using the request context

DasProffi:
closes #296


MaxSchaefer (approved):
Looks good! As discuessed on Slack, we will later propose a common naming convention for endpoints that returns scoped results by some sort of context. We prefer “Self” → "GetSelfOrganizations".

PR #300, opened 2023-07-22
fix: organization is_personal default to false via optional

MaxSchaefer:
closes #


use-to (approved):

MaxSchaefer:
This pull request sets the `is_personal` attribute to optional. When gRPC sends false, it's actually null on the wire. The getter of optional attributes will cover the nil-Check and returns the default value of false.

DasProffi (approved):

PR #302, opened 2023-07-23
add GetInvitationsByOrganization

### LINKED ISSUE #299: `GetInvitationsByOrganization`
### Describe the feature

The response should be the same as [GetInvitationsByUserResponse](https://proto.helpwave.de/#proto.services.user_svc.v1.GetInvitationsByUserResponse) just without the organization. The organization ID gets passed in the request as `organization_id`. Please check via the Claims if the user is able to access the requested organization.

The frontend wants to list all invitations of the organization.

### Additional context

_No response_

DasProffi:
closes #299


services/user-svc/internal/organization/organization.go
```diff
@@ -367,57 +367,63 @@ func (s ServiceServer) InviteMember(ctx context.Context, req *pb.InviteMemberReq
...
-		return nil, status.Error(codes.InvalidArgument, "only pending invitations can be accepted")
+	hasAccess, err := IsInOrganization(db, organizationID, userID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	// Update invitation state
-	invitation := Invitation{
-		ID: invitationId,
+	if !hasAccess || hwutil.Contains(claims.Organizations, organizationID) {
```
MaxSchaefer, line 394: We don't need to do this membership check twice. The result of `hasAccess` comes from the database. `hwutil.Contains(claims.Organizations, organizationID)` comes from ID token. We rely on and prefer the later one.

services/user-svc/internal/organization/organization.go
```diff
@@ -367,57 +367,63 @@ func (s ServiceServer) InviteMember(ctx context.Context, req *pb.InviteMemberReq
...
+	if req.State != nil {
+		filter = filter.Where("state = ?", req.State)
 	}
 
-	// Add user to organization
-	if err := AddUserToOrganization(ctx, db, userID, currentInvitation.OrganizationID); err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+	if err := filter.Find(&invitations).Error; err != nil {
+		if hwgorm.IsOurFault(err) {
+			log.Warn().Err(err).Msg("database error")
```
MaxSchaefer, line 408:
```suggestion
```
Not necessary. Every returned error will be logged by our [middleware](https://github.com/helpwave/services/blob/main/libs/common/grpc.go#L285).

services/user-svc/internal/organization/organization.go
```diff
@@ -684,3 +743,22 @@ func ChangeMembershipAdminStatus(ctx context.Context, db *gorm.DB, userID uuid.U
 
 	return nil
 }
+
+func IsInOrganization(db *gorm.DB, organizationID uuid.UUID, userID uuid.UUID) (bool, error) {
```
MaxSchaefer, line 747:
```suggestion
func IsInOrganization(db *gorm.DB, organizationID uuid.UUID, userID uuid.UUID) (bool, error) {
```
Same as above. The organization ID for the current user is inside the claims.

MaxSchaefer (changes requested):

services/user-svc/internal/organization/organization.go
DasProffi: yes, but do we update the token every time the user joins a new organization, otherwise the user could be part of the organization, but cannot access the data

services/user-svc/internal/organization/organization.go
DasProffi: I just copy pasted it from some other method, we should open an issue to remove those warnings

DasProffi: opened #303

services/user-svc/internal/organization/organization.go
MaxSchaefer: > but do we update the token every time the user joins a new organization
Yes, after while. When the access token expires.

Correct, but this is our consistency model for the access token. When the request hits that method, [the `hwutil.Contains(claims.Organizations, organizationID)` check was already performed](https://github.com/helpwave/services/blob/main/libs/common/grpc.go#L164) with the OrganizationID of the `X-Organization` metadata. The user is likely to call this method for the current organization. Therefore, the user is already a part of the organization and the later condition will always be true.

MaxSchaefer (approved):

PR #304, opened 2023-07-24
add due date and created by to tasks

### LINKED ISSUE #286: Changes to task and subtasks
### Describe the feature

- [ ] Tasks & Subtasks: Add `created_by` uuid. Add The ID gets filled by the API with the current user.
- [ ] Tasks: Add `due_at` timestamp
- [ ] Tasks: Make `due_at` and `public` updatable
- [ ] Tasks & Subtasks: return newly added attributes to `GetTask`, `GetTasksByPatient` and `GetTasksByPatientSortedByStatus`

DasProffi:
closes #286


services/task-svc/migrations/000013_init.up.sql
MaxSchaefer, line 1: We should prefer `NULL` over magic default values.

MaxSchaefer (changes requested):

services/task-svc/migrations/000013_init.up.sql
DasProffi: for the timestamp i agree, but also for the created_by?

DasProffi: But how should this null value be used? 
Do we want to make the timestamp entirely optional? This causes an issue in the protobuf for update, because there is no optional nullable value. Or do you want to implement an endpoint for removing a due date or do we just allow the backend to have an undefined due date and then still send a default value or null and the frontend has no possibility of removing an due_date?

MaxSchaefer: Sorry for the confusion.

The `created_by` should be NOT NULL (as it already is) without a default value, like [in this example](https://github.com/helpwave/services/blob/ca29fd3061e382f5788b8e47f26ba9055b0121bf/services/user-svc/migrations/000001_init.up.sql#L10).

`due_at` is optional for the user, this should be nullable/optional. An `optional google.protobuf.Timestamp` will be a pointer in Golang and therefore nullable.

DasProffi: > The `created_by` should be NOT NULL (as it already is) without a default value, like [in this example](https://github.com/helpwave/services/blob/ca29fd3061e382f5788b8e47f26ba9055b0121bf/services/user-svc/migrations/000001_init.up.sql#L10).

That only works on create, when altering a table `not null` causes the migration to fail

> `due_at` is optional for the user, this should be nullable/optional. An `optional google.protobuf.Timestamp` will be a pointer in Golang and therefore nullable.

Yes but it would be against our update sematics normally omitting a field means no change, here an omitted due_at would mean set it to null

MaxSchaefer: > That only works on create, when altering a table not null causes the migration to fail
Got it. I've tested this behavior.

> Yes but it would be against our update sematics normally omitting a field means no change, here an omitted due_at would mean set it to null
Right now, the timestamp field is optional in the update protos. When omitted, the due_at gets not touched. 

Overall, I understand that the migrations require these default values. I'm not a huge fan of this. But we need to develop a clean concept for Null UUIDs, these are implemented differently across services.

MaxSchaefer (changes requested):
Looks good. Tested it already!
> Tasks & Subtasks: return newly added attributes to GetTask, GetTasksByPatient and GetTasksByPatientSortedByStatus

Could you please return `created_by` for tasks and subtasks on the mentioned methods as well?

PR #312, opened 2023-07-25
update names of migrations

FoseFx:
It was harder to catch up with the migrations than needed, so I introduced proper file names


DasProffi (approved):

MaxSchaefer (approved):

FoseFx:
image

PR #316, opened 2023-07-26
feat: add name to beds

### LINKED ISSUE #315: Add `name` to bed
### Describe the feature

The bed entity should have a `name` attribute. This attribute is required during creation and optional for updates. Every read operation of beds should return the beds in lexicographic order by name.

This request comes out of the medical team in favor for bed index.

### Additional context

Blocks #246 and #269

FoseFx:
closes #315

The following endpoints are impacted, feel free to use the checkboxes:
- [ ] CreateBed
- [ ] GetBed
- [ ] GetBeds
- [ ] GetBedsByRoom
- [ ] UpdateBed
- [ ] GetRoom
- [ ] GetRooms
- [ ] GetRoomsByWard
- [ ] GetRoomOverviewsByWard
- [ ] GetWardDetails


services/task-svc/internal/room/models/models.go
```diff
@@ -23,4 +23,5 @@ type Bed struct {
 	Room           *Room           `gorm:"foreignKey:ID"`
 	Patient        *models.Patient `gorm:"foreignKey:BedID"`
 	OrganizationID uuid.UUID       `gorm:"column:organization_id"`
+	Name           string          `gorm:"column:name"`
 }
```
DasProffi, line 27: the bed model should be in `bed/models/` and not `room/models`

DasProffi (changes requested):

services/task-svc/internal/room/models/models.go
FoseFx: I replaced the bed/models Bed struct with an alias to this room/models Bed struct. There were two Bed structs before. It has to be here because it references Room, and structs in Room reference Bed, which would cause an import cycle, which is disallowed in go. In the future we should refactor this to a single models package.

DasProffi (approved):

PR #337, opened 2023-08-03
feat: optional fallback for X-Organization header

### LINKED ISSUE #336: `X-Organization` header should be optional
We discussed this in our weekly. Due to some refactoring on the multi-tenancy concept, this issue is very likely the first one that deprecates the organization scoped entities per service.
_A tracking issue with a proper concept will follow._

- [ ] Introduce a new environment variable called `ORGANIZATION_ID`, should hold a UUID
- [ ] Header `X-Organization` should be optional
- [ ] If the header is not set, fallback to the environment variable `ORGANIZATION_ID`

MaxSchaefer:
closes #336 

This pull request defines the `X-Organization` header as optional (previous required) for clients. When the header is not passed in the request and a fallback was passed on startup via an environment variable, this env will be used. If both, header and env, are not set, the request fails like before.

### DIFF
--- a/libs/common/grpc.go
+++ b/libs/common/grpc.go
@@ -145,6 +145,7 @@ func authFunc(ctx context.Context) (context.Context, error) {
 // The claims are signed. Therefore, we can match the user provided
 // organizationID from the headers against the organizationIDs inside the claim.
 func handleOrganizationIDForAuthFunc(ctx context.Context) (context.Context, error) {
+	log := zlog.Ctx(ctx)
 	organizationIDStr, err := OrganizationIDFromMD(ctx)
 
 	var organizationID uuid.UUID
@@ -154,13 +155,13 @@ func handleOrganizationIDForAuthFunc(ctx context.Context) (context.Context, erro
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "invalid organizationID")
 		}
-	} else if OrganizationID != nil {
+	} else if InstanceOrganizationID != nil {
 		// no organizationID in header but in envs, using env
-		zlog.Ctx(ctx).Info().Msg("no valid organization header found, falling back to environment organization id")
-		organizationID = *OrganizationID
+		log.Trace().Msg("no valid organization header found, falling back to environment organization id")
+		organizationID = *InstanceOrganizationID
 	} else {
 		// no organizationID in header or env, error
-		zlog.Ctx(ctx).Trace().Err(err).Msg("no valid organization header found")
+		log.Trace().Err(err).Msg("no valid organization header found")
 		return nil, err
 	}
 
@@ -170,15 +171,15 @@ func handleOrganizationIDForAuthFunc(ctx context.Context) (context.Context, erro
 	}
 
 	if !hwutil.Contains(claims.Organizations, organizationID) {
-		zlog.Ctx(ctx).Info().Str("organizationID", organizationID.String()).Msg("organization in header was not part of claims")
+		log.Info().Str("organizationID", organizationID.String()).Msg("organization in header was not part of claims")
 		return nil, status.Errorf(codes.Unauthenticated, "no access to this organization")
 	}
 
 	ctx = context.WithValue(ctx, organizationIDKey{}, organizationID)
 
 	// Append organizationID to the logger
-	log := zlog.Ctx(ctx).With().Str("organizationID", organizationID.String()).Logger()
-	return log.WithContext(ctx), nil
+	loggerWithOrganizationID := zlog.Ctx(ctx).With().Str("organizationID", organizationID.String()).Logger()
+	return loggerWithOrganizationID.WithContext(ctx), nil
 }
 
 // VerifyFakeToken accepts a Base64 encoded json structure with the schema of IDTokenClaims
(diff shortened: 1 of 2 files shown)


libs/common/setup.go
```diff
@@ -16,6 +17,7 @@ import (
 var (
 	Mode                    string // Mode is set in Setup()
 	InsecureFakeTokenEnable = false
+	OrganizationID          *uuid.UUID
```
FoseFx, line 20:
```suggestion
	InstanceOrganizationID          *uuid.UUID
```
or something else, but just `OrganizationID` is a) confusing and b) will cause a great deal of mismatch bugs with `organizationID` variables

libs/common/grpc.go
```diff
@@ -146,16 +146,24 @@ func authFunc(ctx context.Context) (context.Context, error) {
...
+	var organizationID uuid.UUID
+	if err == nil {
+		// organizationID in header, using header
+		organizationID, err = uuid.Parse(organizationIDStr)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "invalid organizationID")
+		}
+	} else if OrganizationID != nil {
+		// no organizationID in header but in envs, using env
+		zlog.Ctx(ctx).Info().Msg("no valid organization header found, falling back to environment organization id")
```
FoseFx, line 159:
```suggestion
		zlog.Ctx(ctx).Trace().Msg("no valid organization header found, falling back to environment organization id")
```

FoseFx: also there are now two instances where a logger is used, so we should declare a log variable at the top of the function instead

libs/common/setup.go
```diff
@@ -58,6 +60,16 @@ func Setup(serviceName, version string, auth bool) {
 
 		setupAuth()
 	}
+
+	organizationIdStr := hwutil.GetEnvOr("ORGANIZATION_ID", "")
```
FoseFx, line 64: rename these variables consistent with the global variable (I suggested it above somewhere)

FoseFx: maybe add a comment why this variable exists and when it is used

libs/common/setup.go
```diff
@@ -58,6 +60,16 @@ func Setup(serviceName, version string, auth bool) {
...
+	if organizationIdStr != "" {
+		organizationID, err := uuid.Parse(organizationIdStr)
+		if err != nil {
+			log.Fatal().Err(err).Msg("invalid uuid for environment variable ORGANIZATION_ID")
+		}
+		log.Info().Str("organizationID", organizationID.String()).Msg("using fallback organizationID")
```
FoseFx, line 70:
```suggestion
		log.Info().Str("organizationID", organizationID.String()).Msg("using default organizationID")
```
it's not really a fallback, we expect the header to not be set in a deployment that uses it

MaxSchaefer: When the `X-Organization` header is not set, we are trying to _fallback_ to this environment variable. I've added a comment anyway around this code.

libs/common/setup.go
```diff
@@ -56,6 +58,18 @@ func Setup(serviceName, version string, auth bool) {
...
 
+		// organizationIdStr, later InstanceOrganizationID is used as a fallback when a client does not send the organization header
+		// For code consistency purposes, we are parsing organizationIdStr from a string into a UUID
+		organizationIdStr := hwutil.GetEnvOr("ORGANIZATION_ID", "")
+		if organizationIdStr != "" {
+			organizationID, err := uuid.Parse(organizationIdStr)
+			if err != nil {
+				log.Fatal().Err(err).Msg("invalid uuid for environment variable ORGANIZATION_ID")
+			}
+			log.Info().Str("organizationID", organizationID.String()).Msg("specified fallback organizationID when no organization header was found")
```
FoseFx, line 69:
```suggestion
			log.Info().Str("organizationID", organizationID.String()).Msg("specified fallback organizationID for requests without organization header")
```

libs/common/grpc.go
```diff
@@ -162,15 +171,15 @@ func handleOrganizationIDForAuthFunc(ctx context.Context) (context.Context, erro
...
+		log.Info().Str("organizationID", organizationID.String()).Msg("organization in header was not part of claims")
 		return nil, status.Errorf(codes.Unauthenticated, "no access to this organization")
 	}
 
 	ctx = context.WithValue(ctx, organizationIDKey{}, organizationID)
 
 	// Append organizationID to the logger
-	log := zlog.Ctx(ctx).With().Str("organizationID", organizationID.String()).Logger()
-	return log.WithContext(ctx), nil
+	loggerWithOrganizationID := zlog.Ctx(ctx).With().Str("organizationID", organizationID.String()).Logger()
```
FoseFx, line 181: might need some attention

FoseFx (approved):

libs/common/grpc.go
MaxSchaefer: Can you elaborate on that? The code is not _very nice_, but it's the same as above for the `userID`.

FoseFx: could have been just
```go
log = log.With().Str("organizationID", organizationID.String()).Logger()
return log.WithContext(ctx), nil
```

PR #343, opened 2023-08-06
ci: fix unauthed deploys

### LINKED ISSUE #339: Pushing to new branch without PR causes deployment CI to run
### Describe the bug

We need to check if we are on main in checks like the following:
https://github.com/helpwave/services/blob/main/.github/workflows/cicd.yaml#L102

1. Create a new branch a push changes to it
2. Don't create a PR for it
3. Change is deployed

### Additional context

https://github.com/helpwave/services/actions/runs/5762996167/job/15623902760

FoseFx:
closes #339

### DIFF
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -42,6 +42,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ github.token }}
 # TODO (arm-versioning): Arm versioning by removing the comments of the following lines
 #      - uses: bufbuild/buf-breaking-action@v1
 #        with:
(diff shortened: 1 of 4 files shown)


FoseFx:
I know what you're thinking, and I agree. I would also love to store this in a variable for the whole workflow, but, afaik, you can only define environment variables, not github(?) variables, for a workflow. But a job is able to overwrite this. If a malicious actor writes go code to modify the environment, they might be able to force a deployment.

MaxSchaefer:
@FoseFx any news?

FoseFx:
I removed the env again and now it works. I'd leave it as it is, if you want to open a follow-up issue to improve it, feel free to do so.

.github/workflows/buf.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ github.token }}
```
MaxSchaefer, line 46: Reason?

MaxSchaefer (approved):

.github/workflows/buf.yaml
FoseFx: The wf was flaky due to rate-limiting (unauthed requests)

PR #348, opened 2023-08-06
add GetAssignedTasks

### LINKED ISSUE #338: Add `GetAssignedTasks`
This endpoint `GetAssignedTasks` should return the assigned tasks of the user requesting the method. Make sure to abstract the actual query behind a method that returns the tasks and accepts a userID.

The response should be the same as [GetTasksByPatientResponse](https://proto.helpwave.de/#proto.services.task_svc.v1.GetTasksByPatientResponse). Adjust `patient_id` as follows.
```
{
  ...
  "patient": {
    "id": string,
    "name": string
  },
  ...
}
```

DasProffi:
closes #338


proto/services/task_svc/v1/task_svc.proto
```diff
@@ -124,6 +124,42 @@ message GetTasksByPatientSortedByStatusResponse {
 }
 
 
+//
+// GetAssignedTasks
+//
+
+message GetAssignedTasksRequest {
+}
```
MaxSchaefer, line 132:
```suggestion
message GetAssignedTasksRequest {}
```

services/task-svc/internal/task/task.go
```diff
@@ -249,6 +250,66 @@ func (ServiceServer) GetTasksByPatientSortedByStatus(ctx context.Context, req *p
...
+	}
+
+	var tasks []*pb.GetAssignedTasksResponse_Task
+
+	for _, patient := range patients {
+		mappedPatient := &pb.GetAssignedTasksResponse_Task_Patient{
+			Id:   patient.ID.String(),
+			Name: patient.HumanReadableIdentifier,
+		}
+		fmt.Println(len(patient.Tasks))
```
MaxSchaefer, line 283:
```suggestion
```

MaxSchaefer (changes requested):
> Make sure to abstract the actual query behind a method that returns the tasks and accepts a userID.

The actual mapping should then be done at the gRPC-method level.

MaxSchaefer (approved):

PR #351, opened 2023-08-10
make organisation id optional in GetPatient list

### LINKED ISSUE #341: GetPatientList Organization ID not optional
### Describe the bug

```proto
message GetPatientListRequest {
  string organisation_id = 1;
}
```

try to get the patientlist with the default organisation (no organization id in request)

### Additional context

_No response_

DasProffi:
closes #341


services/task-svc/internal/repositories/patient_repository.go
```diff
@@ -83,7 +83,8 @@ func (r *PatientRepository) GetUnassignedPatientsForOrganization(organizationID
 func (r *PatientRepository) GetDischargedPatientsForOrganization(organizationID uuid.UUID) ([]models.Patient, error) {
 	var patients []models.Patient
 	query := r.db.
-		Where("organization_id = ? AND is_discharged = 1", organizationID).
+		Unscoped().
+		Where("organization_id = ? AND NOT is_discharged = 0", organizationID).
```
MaxSchaefer, line 87: Why did you negate this condition, and why is this change in the scope of this pull request?

MaxSchaefer (changes requested):

services/task-svc/internal/repositories/patient_repository.go
DasProffi: because the request didn't return any discharged patients, because we need the unscoped to see the soft deleted entries.
The `NOT is_discharged = 0` is there, because the softdelete is a timestamp and any value that isn't the default (not deleted) indicates deletion. Thus checking `discharged = 0` might not work properly

MaxSchaefer (approved):

PR #357, opened 2023-08-14
feat: first draft for deleting patient

### LINKED ISSUE #330: PatientDelete
### Describe the feature

Delete Patient request only using the patient id

because we use soft deletes from gorm, but we also need the function to delete all patient information

### Additional context

_No response_

florianjs:
closes #330


DasProffi:
Looks great, only the admin check is still missing

florianjs:
> Looks great, only the admin check is still missing

I wanted to avoid implementing the admin check twice in user-svc and task-svc. We should implement this in the longer term in the authclaims

DasProffi (approved):

DasProffi:
> > Looks great, only the admin check is still missing
> 
> I wanted to avoid implementing the admin check twice in user-svc and task-svc. We should implement this in the longer term in the authclaims

Makes sense, especially since we don't have the information in the task service to validate this

PR #362, opened 2023-08-16
fix: make proto

florianjs:
fix pipeline


florianjs:
Looks like the pipeline is failing because changes were added after my  Pull Request causing problems with the protbuf. Unfortunately my pull request is already merged so I want to fix the error with this pull request.

FoseFx (approved):
How did this happen?

PR #365, opened 2023-08-19
add LRU module

### LINKED ISSUE #335: Implement an internal abstraction for recent entities
The goal is to have an internal abstraction / function / service available to publish keys with a lifetime X into our redis that holds Y amount of identifiers for a group of entities. The map should also keep a maximum of Z identifiers (Ring Buffer, First-In First-Out).

Example
```
addRecentEntity('wards', userId, wardId); // appends the wardId to a map that is scoped to a user
getRecentEntities('wards', userId) // returns the recent wardIds for a user
```

FoseFx:
closes #335


libs/decaying_lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+		Stop:  size - 1,
+		Rev:   true,
+	})
+
+	return res.Result()
+}
+
+// AddItem adds an item for your key for a user,
+// if you don't want to store information scoped to a user see AddItemForKey
+func (lru *DecayingLRU) AddItem(key, userID, item string) error {
```
MaxSchaefer, line 93:
```suggestion
func (lru *DecayingLRU) AddItemForUser(key, userID, item string) error {
```

libs/decaying_lru/lru.go
```diff
@@ -0,0 +1,101 @@
...
+	return lru.AddItemForKey(key+"-"+userID, item)
+}
+
+// GetItems gets items for your key for a user,
+// if you don't want to access information scoped to a user see GetItemsForKey
+func (lru *DecayingLRU) GetItems(key, userID string) ([]string, error) {
```
MaxSchaefer, line 99:
```suggestion
func (lru *DecayingLRU) GetItemsForUser(key, userID string) ([]string, error) {
```

MaxSchaefer (approved):

MaxSchaefer:
@FoseFx I've added `RemoveItemByValue()` to remove items directly from the LRU. For example, when a tracked entity gets removed.

PR #371, opened 2023-08-21
recent patients tracking

### LINKED ISSUE #367: Track recent patients per user
With #365 in place, we can build up a record of recent patients per user.

Every following method triggers a recent interaction with a patient
- `CreatePatient`
- `UpdatePatient`
- `GetPatientDetails`

Every following method should remove the patient from the LRU 
- `DischargePatient`
- `DeletePatient`

FoseFx:
closes #367

### DIFF
--- a/services/task-svc/internal/patient/patient.go
+++ b/services/task-svc/internal/patient/patient.go
@@ -368,6 +368,7 @@ func (ServiceServer) GetPatientDetails(ctx context.Context, req *pb.GetPatientDe
 		}
 	})
 
+	// TODO: check if tracking here makes sense or too much spam
 	tracking.AddPatientToRecentActivity(ctx, id.String())
 
 	return &pb.GetPatientDetailsResponse{
--- a/services/task-svc/internal/patient/patient.go
+++ b/services/task-svc/internal/patient/patient.go
@@ -13,6 +13,7 @@ import (
 	pbhelpers "proto_helpers/task_svc/v1"
 	"task-svc/internal/models"
 	"task-svc/internal/repositories"
+	"task-svc/internal/tracking"
 )
 
 type ServiceServer struct {
@@ -49,6 +50,8 @@ func (ServiceServer) CreatePatient(ctx context.Context, req *pb.CreatePatientReq
 	log.Info().
 		Str("patientId", patient.ID.String()).
 		Msg("patient created")
+
+	tracking.AddPatientToRecentActivity(ctx, patient.ID.String())
 
 	return &pb.CreatePatientResponse{
 		Id: patient.ID.String(),
@@ -206,6 +209,8 @@ func (ServiceServer) UpdatePatient(ctx context.Context, req *pb.UpdatePatientReq
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	tracking.AddPatientToRecentActivity(ctx, id.String())
+
 	return &pb.UpdatePatientResponse{}, nil
 }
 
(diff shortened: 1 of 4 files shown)


services/task-svc/internal/tracking/tracking.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	PatientKey = "recently-viewed-patient"
+)
+
+var lru decaying_lru.DecayingLRU
+
+func SetupTracking(serviceName string) {
+	lru = decaying_lru.Setup(serviceName, 5, 4*24*time.Hour, 10)
```
use-to, line 18: Parameterize this, please. The default delay should be defined elsewhere.

services/task-svc/internal/tracking/tracking.go
```diff
@@ -0,0 +1,51 @@
...
+
+func getUserID(ctx context.Context) string {
+	log := zlog.Ctx(ctx)
+
+	userID, err := common.GetUserID(ctx)
+
+	if err != nil {
+		log.Error().Err(err).Msg("could not get userID from context, can't add item to LRU")
+		return ""
+	}
```
use-to, line 29: This error handling make more sense inside the `AddPatientToRecentActivity` function.

services/task-svc/internal/patient/patient.go
```diff
@@ -361,6 +368,8 @@ func (ServiceServer) GetPatientDetails(ctx context.Context, req *pb.GetPatientDe
 		}
 	})
 
+	tracking.AddPatientToRecentActivity(ctx, id.String())
```
use-to, line 371: Add this patient to recent activity in order to suggest a shorter patient list is pretty complicated, as this function will be requested pretty often without any intention of operating or managing that patient.

So I would argue that this is no necessarily the best place to detect recent activity, but that's discussable (we could try this).
But definitely add `AssignBed` and `UnassignBed` to the "detection list".

use-to (approved):

use-to (changes requested):
Do not merge.

use-to:
@MaxSchaefer What the current status of our Redis instance? What is the priority of this problem?

MaxSchaefer (changes requested):
We are already having the required Redis environment variables through Dapr. You could retrieve those secrets via https://docs.dapr.io/developing-applications/building-blocks/secrets/howto-secrets/#calling-the-secrets-api-from-your-code. The secretstore is definied here https://github.com/helpwave/services/blob/main/dapr/components/env-secret-store.yaml. And these are our envs that are fetched by the secretstore:
SECRETSTORE_REDIS_HOST, SECRETSTORE_REDIS_PASSWORD, SECRETSTORE_REDIS_USERNAME.

But for now, you can set the LRU up via the environment variables directly.

services/task-svc/internal/tracking/tracking.go
FoseFx: No, it will be required in `AddPatientToRecentActivity`, `RemovePatientFromRecentActivity`, `AddWardToRecentActivity` and `RemoveWardFromRecentActivity`, possibly even more. I don't want to repeat this everywhere

FoseFx:
Does this look right? @MaxSchaefer

FoseFx:
also let me know if `INSECURE_REDIS_NO_TLS` needs to be set in staging

use-to (approved):

MaxSchaefer (approved):

PR #374, opened 2023-08-23
feat: create user and organization on registration

MaxSchaefer:
This pull requests is the first version of our onboarind workflow for new accounts. When Ory send the after registration webhook to our API, we create a new account and organization. The issued token right after the registration holds a default organization (due to our current multi-tenancy limitations, to-do for later on). After another sign-in the token gets correctly populated with the additional private personal organization.

Register a new account on https://tasks.helpwave.de


libs/common/setup.go
```diff
@@ -23,9 +23,11 @@ var (
 const DevelopmentMode = "development"
 const ProductionMode = "production"
 
+var skipAuthForMethods []string
+
 // Setup loads the .env file and sets up logging
 // also sets up tokens when the service requires auth
-func Setup(serviceName, version string, auth bool) {
+func Setup(serviceName, version string, auth bool, unauthenticatedMethods *[]string) {
```
use-to, line 30: Create overloading function in order to prevent appended `nil` passing on every call.

libs/common/setup.go
```diff
@@ -23,9 +23,11 @@ var (
 const DevelopmentMode = "development"
 const ProductionMode = "production"
 
+var skipAuthForMethods []string
```
use-to, line 26: The naming could be better or more related to the parameter passed in `Setup()`.

services/ory-svc/main.go
```diff
@@ -86,14 +103,59 @@ func afterRegistrationWebhookHandler(ctx context.Context, in *daprcmn.Invocation
...
 		Email:    payload.Email,
 		Nickname: payload.Nickname,
 		Name:     payload.Name,
+	})
+
+	if err != nil {
+		return nil, newErrAndLog(ctx, err.Error())
+	}
+
+	trueVal := true
```
use-to, line 127: Really? Maybe we should create a helper function for passing a boolean pointer based on a fixed value.

use-to (changes requested):

use-to (approved):
As my requested changes are low priority for now. Feel free to merge on at your own discretion.

PR #403, opened 2023-09-10
feat: sort subtasks by creation date

### LINKED ISSUE #402: Sort subtasks
### Describe the feature

We should sort the subtasks directly on the backend to avoid some site effects, see [#726](https://github.com/helpwave/web/issues/726)

To avoid some side effects on the frontend

### Additional context

_No response_

florianjs:
closes #402

### DIFF
--- a/services/task-svc/migrations/000017_creation_date_subtask.up.sql
+++ b/services/task-svc/migrations/000017_creation_date_subtask.up.sql
@@ -1,0 +1,3 @@
+-- add constraint
+ALTER TABLE subtasks
+	ADD COLUMN creation_date timestamp DEFAULT CURRENT_TIMESTAMP;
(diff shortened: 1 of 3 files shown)


services/task-svc/migrations/000017_creation_date_subtask.up.sql
```diff
@@ -0,0 +1,3 @@
+-- add constraint
```
FoseFx, line 1: no constraint, also see down

FoseFx (approved):

services/task-svc/migrations/000017_creation_date_subtask.up.sql
florianjs: You are very fast 🚀

PR #426, opened 2023-09-22
Issue/423 impulse svc

Paul-B98:
closes #


services/impulse_svc/Dockerfile
```diff
@@ -0,0 +1,10 @@
+From python:3.11.5-alpine
```
MaxSchaefer, line 1:
```suggestion
FROM python:3.11.5-alpine
```

MaxSchaefer (changes requested):

MaxSchaefer (approved):

PR #453, opened 2023-09-25
fix: CreateUser, UpdateUser, GetScore, GetRewards, TrackChallenge, St…

### LINKED ISSUE #452: Refactor grpc endpoints (impulse-svc)
### Describe the bug

Improve request validation and return better error msg's.

Send requests without userId to functions.

### Additional context

_No response_

Friedjof:
…atsForTeamByUser and Verification

closes #452 

Send grpc requests.

### DIFF
--- a/services/impulse_svc/grpc_service/service.py
+++ b/services/impulse_svc/grpc_service/service.py
@@ -1,12 +1,12 @@
 import sys
-
 from datetime import datetime
+from uuid import UUID
 
 import django.core.exceptions as exceptions
 from django.utils.timezone import make_aware
+from django.db.models import Q
 
 import grpc
-from uuid import UUID
 
 sys.path.append("./gen/")
 
@@ -15,23 +15,19 @@ from proto.services.impulse_svc.v1 import impulse_svc_pb2
 
 from grpc_service.models import Challenge, UserChallenge, User, Reward, Team
 
-from django.db.models import Q
 
-
-def is_uuid(uuid_input: str, version=4) -> bool:
+def get_uuid(uuid_input: str, version=4) -> UUID | None:
     try:
-        UUID(uuid_input, version=version)
+        return UUID(uuid_input, version=version)
     except ValueError:
-        return False
-    return True
+        return None
 
 
-def is_iso_datetime(datetime_input: str) -> bool:
+def get_iso_datetime(datetime_input: str) -> datetime | None:
     try:
-        datetime.fromisoformat(datetime_input)
+        return datetime.fromisoformat(datetime_input)
     except ValueError:
-        return False
-    return True
+        return None
 
 
 def parse_datetime(datetime_input: str) -> datetime:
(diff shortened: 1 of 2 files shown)


Friedjof:
More values could be checked, such as the pal value or the score in TrackChallenge.

MaxSchaefer:
Nice! Is there any Django build-in validation framework available? In go we are [utilizing the comments in our Protobufs](https://github.com/helpwave/services/blob/main/proto/services/task_svc/v1/task_svc.proto#L20).

@Paul-B98

Friedjof:
The basic validation is already done in the grcp plugin for django like types or if values are present. However, types such as uuid can not be regulated to my knowledge via grcp and must be passed as a string. String to date can certainly be done in Djabgo, but for that I have the standard libs directly. Other data like negative lengths I check then by a plausibility check. We could discuss again how to formulate this more precisely (e.g. the pal value is not greater than 3.0 as far as I know). I tried to do without try except, but this is not always possible. In django there are validators for rest or formilare. gRPC brings there as said already so eimiges with. The rest I could just validate by the current checks. Maybe we will switch to go in the future.

services/impulse_svc/grpc_service/service.py
```diff
@@ -3,9 +3,10 @@
 from datetime import datetime
 
 import django.core.exceptions as exceptions
+from django.utils.timezone import make_aware
 
-import pytz
 import grpc
+from uuid import UUID
```
Paul-B98, line 9: Imports should be ordered in PEP8 style

services/impulse_svc/grpc_service/service.py
```diff
@@ -15,14 +16,45 @@
...
+        return False
+    return True
+
+
+def is_iso_datetime(datetime_input: str) -> bool:
+    try:
+        datetime.fromisoformat(datetime_input)
+    except ValueError:
+        return False
+    return True
```
Paul-B98, line 34: How about returning a tuple with the parsed value or None?

services/impulse_svc/grpc_service/service.py
```diff
@@ -35,47 +67,60 @@ def CreateUser(self, request: impulse_svc_pb2.CreateUserRequest, context):
...
-            if request.birthday != "":
-                context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
-                context.set_details("Check the date format")
-                return impulse_svc_pb2.UpdateUserResponse()
+        if User.objects.filter(id=request.user_id).count() == 0:
+            context.set_code(grpc.StatusCode.NOT_FOUND)
+            context.set_details("user does not exist")
+            return impulse_svc_pb2.UpdateUserResponse()
+
+        user: User = User.objects.get(id=request.user_id)
```
Paul-B98, line 89: I think a try except with `User.DoesNotExists` exception could be more suitable

Paul-B98 (changes requested):

services/impulse_svc/grpc_service/service.py
Friedjof: Do you mean something like this?
```python
def get_uuid(uuid_input: str, version=4) -> UUID | None:
  pass

def get_iso_datetime(datetime_input: str) -> datetime | None:
  pass
```

Friedjof:
I have now pushed the changes. I also fixed one thing with user_id in the verification. I don't know exactly why user_id had crept in there instead of challenge_id.
I have tested all basic functions and found nothing more. There is still one TODO in the code regarding the verification of the PAL value. We could define more precisely in which range it may lie.

use-to:
@Friedjof Once you have completed your review thread or implemented the necessary changes, please click on the "Resolve conversation" button.
@Paul-B98 could you please re-review this pr?

MaxSchaefer:
@Friedjof there was no outstanding review or conversation left. I merged this pr. Thanks for the refactor!

Issue #458, opened 2023-10-05
No storing of recent patients with no error on Prod/Staging Redis


MaxSchaefer:
### Describe the bug

The `decay_lru` is not complaining about any problems with redis and is not storing anything in redis. This problem only occurs on Prod/Staging.

### Describe how to produce the bug

Using `UpdatePatient` and `GetRecentPatients` on Prod/Staging

### Additional context

https://github.com/helpwave/web/pull/744

FoseFx:
I replace the upstash instance with a temporary "real" redis instance, and it works fine.

<details>

<summary> Some Logs </summary>

```
127.0.0.1:6379> MONITOR
OK
1697716398.310680 [0 [fdaa:2:44d0:a7b:141:4d88:57df:2]:44806] "evalsha" "7a1c3c4f4a4ef54db45c48e5453544d66de490fc" "1" "recently-viewed-patient-18159713-5d4e-4ad5-94ad-fbb6bb147984" "ac10b72f-5b55-4a5f-b39e-3bd717166e27" "10" "86400000000000" "20"
1697716398.310746 [0 lua] "TIME"
1697716398.310762 [0 lua] "ZADD" "recently-viewed-patient-18159713-5d4e-4ad5-94ad-fbb6bb147984" "1697716398" "ac10b72f-5b55-4a5f-b39e-3bd717166e27"
1697716398.310783 [0 lua] "EXPIRE" "recently-viewed-patient-18159713-5d4e-4ad5-94ad-fbb6bb147984" "86400000000000"
1697716398.315802 [0 [fdaa:2:44d0:a7b:141:4d88:57df:2]:35858] "hello" "3" "auth" "(redacted)" "(redacted)"
1697716398.317749 [0 [fdaa:2:44d0:a7b:141:4d88:57df:2]:35858] "evalsha" "7a1c3c4f4a4ef54db45c48e5453544d66de490fc" "1" "recently-viewed-ward-18159713-5d4e-4ad5-94ad-fbb6bb147984" "f7f56ab1-2ec5-4df3-8e55-45a664cd8174" "10" "86400000000000" "20"
1697716398.317794 [0 lua] "TIME"
1697716398.317808 [0 lua] "ZADD" "recently-viewed-ward-18159713-5d4e-4ad5-94ad-fbb6bb147984" "1697716398" "f7f56ab1-2ec5-4df3-8e55-45a664cd8174"
[...]

PR #482, opened 2023-10-21
fix: workaround for GetRecentPatients

MaxSchaefer:
This pull request introduces a workaround until https://github.com/helpwave/services/issues/458 is fixed.  If the error occurs, empty response from Redis, we are falling back to the introduced `updated_at` timestamp.


services/task-svc/internal/patient/patient.go
```diff
@@ -204,6 +209,16 @@ func (ServiceServer) GetRecentPatients(ctx context.Context, req *pb.GetRecentPat
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	// WORKAROUND: Until https://github.com/helpwave/services/issues/458 is fixed
+	if len(recentPatientIdsStrs) == 0 {
+		log.Debug().Msg("recentPatientIdsStrs was empty")
+		if patients, err := patientRepo.GetLastUpdatedPatientsForOrganization(4, organizationID); err == nil {
```
florian-sabonchi, line 215: I think we could add a separate variable for reuse

MaxSchaefer: The `4`, not reused right now. Its also just a "`WORKAROUND`".

PR #489, opened 2023-10-25
chore: python prototype service

### LINKED ISSUE #487: Python prototype service
### Describe the feature

This issue tackles the first experimental implementation of a full-functional service in Python.

The service must have a clean software architecture:
- Models represents the database
- Repositories are interacting with models
- Endpoints accessing repositories
- Endpoints are part of the API layer, may be used by different protocols

Beside the architecture, the service needs to full-fill the following requirements:
-  Is able to implement and using generated gRPC services and Protocol Buffers of `gen/`
  - For service-to-service calls
    - All service-to-service calls will go through Dapr gRPC, keep that in mind.
  - Events  
    - Reacting on and Dispatching events
    - Also via Dapr
- Using GraphQL
  - For public-facing APIs
  - Mostly via a federated approach in the future
- Authentication via OIDC, see our Go `libs/common` library for a reference

To-dos:

_One step and adopting at a time_

- [ ] Setup and bootstrapping the service on an `experimental/python-service` branch
  - Django was in the conversation, propose something
  - If we want to use Django "just" for the convenience of an ORM, and otherwise we need to modify / develop workarounds for gRPC and GraphQL, we need to reconsider Django
- [ ] Adopting gRPC on the API layer first
- [ ] Basic structure for the already mentioned software architecture
- [ ] Adopting GraphQL
- [ ] Authentication for both API layers
  - We will probably have to take some shortcuts for internal service-to-service calls, due to the fact that other services right now have just one unified public/private API layer
- [ ] Finalize the architecture

After each step, we should go into a review.

Adopting Python as a language, beside Golang, is a long-term goal. This issue tackles the first experimental implementation of a full-functional service in Python.

### Additional context

- I (@MaxSchaefer) am not sure how Dapr reacts to internal gRPC and "public" GraphQL APIs.
- No business logic is planned yet. If you need to implement against one, you can "clone" the `impulse-svc`

_No response_

Paul-B98:
closes #487


Paul-B98:
I will add a GraphQL example in the near future. At the moment Strawberry is my preferred lib: https://strawberry.rocks/docs/integrations/django#async-django

libs/python/README.md
MaxSchaefer, line 1: Empty. Write or remove.

libs/python/django_grpc/README.md
MaxSchaefer, line 1: Empty. Write or remove.

MaxSchaefer (changes requested):
- Can we get rid of the `services/impulse_svc/gen` directory?
- It seams like that `libs/python` handles djano so please rename this libe to to something like `libs/hwdjano`
- Revisit the empty READMEs

Paul-B98:
> * Can we get rid of the `services/impulse_svc/gen` directory?

Should be possible... I just have to find a goof way to package the gen code correctly.

> * It seams like that `libs/python` handles djano so please rename this libe to to something like `libs/hwdjano`

This was the result of our conversation where we made the decision that we create a language folder structure under lib like the `gen` folder.

MaxSchaefer (changes requested):
- [x] Green CI
- [x] Functional Docker Build of `impulse_svc`

MaxSchaefer:
@Paul-B98 green ci?

MaxSchaefer (approved):

PR #478, opened 2023-10-17
chore: add devenv

MaxSchaefer:
This pull requests tackles DX problems. Especially our initial onboarding.
__To all reviewers: Please try this out without further explanation of my side, to end-to-end test the onboarding and give me feedback.__ The broader idea is already outlined in the issue, before rolling this out organization wide, we will discuss this at the latest in the next weekly. Our old workflow via Makefile is still functional so that the migration can be done over time.

## Which issues does this pull request close?
closes #477 

Open To-dos for later implementation
- I've used [air](https://github.com/cosmtrek/air) for hot-reloading the go services, seams like a proper fit. Maybe we are finding a non-go branded solution for this. There are many out there.
- Doing Dapr component registration so that Dapr finds the redis instance, for example. This is not required yet for every local development.

## [OPTIONAL] Give testing instructions to reviewers

Going through our new _Getting started_ guide!

PR #495, opened 2023-11-06
fix: catch up on ci errors of #478

MaxSchaefer:


libs/common/setup.go
```diff
@@ -90,9 +90,15 @@ func SetupWithUnauthenticatedMethods(serviceName, version string, auth bool, una
 // build the address that can be used by an HTTP or gRPC server.
 // The address will always be in the "(ADDR):PORT" format.
 // If "ADDR" is not set, the "ADDR" part of the format will be "".
-// If "APP_PORT" (through Dapr) is not set try "PORT", the fallback is "8080".
+// If "PORT" is not set, the fallback is "8080".
+// If "APP_PORT" is set, we assume that Dapr is running on the same interface so we return "127.0.0.1:$APP_PORT"
```
FoseFx, line 94: Ok, but why are we doing this? Can you open a follow-up with a comment?

PR #545, opened 2023-11-16
fix: update and tidy common and svc deps

MaxSchaefer:
This pull requests updates and tidying the gomod of common, task-svc and user-svc


FoseFx (approved):
btw the grcp middleware plugin has a v2 now

PR #554, opened 2023-11-20
Create a go.work file

FoseFx:
This PR creates a go workspace, which removes the need for replaces and imports in go.mod files and allows to sync dependencies using "go work sync".


FoseFx:
[Looks like this breaks go get](https://groups.google.com/g/golang-nuts/c/2Ht4c-eZzgQ). Go mod is definitely one of the package managers of our time.

Issue #42, opened 2023-01-17
Migrations in final image


MaxSchaefer:
### Is your feature request related to a problem? Please describe:
> Please provide a clear and concise description of what the problem is.

Right now, there are now migration files inside the final image. 

### Describe the solution you'd like
> Please provide a clear and concise description of what you want to happen.

When we want to run migrations as an initContainer we should include the migration files in the final image.

### Additional Comments
> Add any other context about the feature request here.

We discussed the process to run the migrations out of an initContainer in a meeting today.

PR #582, opened 2023-12-10
feat: kick off eventstore

MaxSchaefer:
This pull request introduces the `hwes` lib. This lib will provide the necessary tooling around [EventStoreDB](https://www.eventstore.com/) for our upcoming event sourcing. Until the rewrite of some entites to event sourced ones, we dispatching some events to the event store as seen in `task-svc`. These features are behind the feature flag `FEATURE_FLAG_EVENT_SOURCING`.

## Give testing instructions to reviewers

- Start EventStoreDB `docker-compose --file dev/docker-compose.yaml up --detach eventstore`
- Run the service with the new `EVENTSTORE_CS` and `FEATURE_FLAG_EVENT_SOURCING` of the `.env.example` file
- Create a bed for example or using the other endpoints
- Open the EventStoreDB UI on localhost:2113 and taking a look at the streams tab


MaxSchaefer:
[ ] Handle if-feature-flag in service instead of lib

MaxSchaefer:
@FoseFx @PaulKalho I've added proper testing instructions

PaulKalho (changes requested):
Looks good, had some problems with the deleteBed Event and the xAssigned Events though.

services/task-svc/internal/events/events.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	log := zlog.Ctx(ctx)
+	err := hwes.AppendEvents(ctx, bedId.String())
+	if err != nil {
+		log.Error().
+			Err(err).
+			Str("event", BedDeleted).
+			Msg("could not dispatch event")
+	}
```
PaulKalho, line 184: You are calling `AppendEvents` without passing an event, is that intended?
The Event is also not showing up in my eventstore db stream

services/task-svc/internal/task/task.go
```diff
@@ -578,6 +579,14 @@ func (ServiceServer) AssignTaskToUser(ctx context.Context, req *pb.AssignTaskToU
...
 		Msg("user assigned")
 
+	if currentUserId, err := common.GetUserID(ctx); err != nil {
+		if currentUserId == userId {
+			_ = events.DispatchTaskSelfAssignedEvent(ctx, id, userId)
+		} else {
+			_ = events.DispatchTaskAssignedEvent(ctx, id, userId)
+		}
+	}
+
```
PaulKalho, line 589: You are dispatching the events in the error handler here

PaulKalho (changes requested):

services/task-svc/internal/events/events.go
MaxSchaefer: Not intended! Thanks.

services/task-svc/internal/task/task.go
MaxSchaefer: Good catch!

MaxSchaefer:
Blocked by #589

libs/hwes/eventstore.go
```diff
@@ -0,0 +1,92 @@
...
+			ctx,
+			id,
+			esdb.AppendToStreamOptions{ExpectedRevision: esdb.NoStream{}},
+			eventsData...,
+		)
+
+		return err
+	}
+
+	lastEvent, err := stream.Recv()
```
PaulKalho, line 65: I discovered that `lastEvent` is the first, not the last event of the stream when I want to dispatch the `BED_DELETEDv1` event. (after dispatching `BED_UPDATEDv1`)

This is probably because `readOps` points into the wrong direction? With line 42 as `readOps` instead, it works like a charm.

libs/hwes/eventstore.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+
+	expectedRevision := esdb.Revision(lastEvent.OriginalEvent().EventNumber)
+
+	_, err = es.AppendToStream(
+		ctx,
+		id,
+		esdb.AppendToStreamOptions{ExpectedRevision: expectedRevision},
```
PaulKalho, line 75: I'm getting a `WrongExpectedStreamRevision` here on `BED_UPDATED_v1` & `BED_DELETED_v1`.

The `BED_UPDATED_v1` is still beeing dispatched and shows up in the db, the `BED_DELETED-v1` not.

PaulKalho (changes requested):

libs/hwes/eventstore.go
MaxSchaefer: Oh yes! The line under the comment in #42 was for testing.

libs/hwes/eventstore.go
MaxSchaefer: I've applied some events including finally the `BED_DELETED_v1`, never got this error. Can you check you local setup? If this error still occurs we can hop on a call :)

PaulKalho: has been fixed by [1ad4583](https://github.com/helpwave/services/pull/582/commits/1ad45834882a80046221afe9cf0d48f2e6bb27d9)

services/task-svc/internal/events/events.go
```diff
@@ -0,0 +1,254 @@
...
+		return hwes.Event{}, err
+	}
+	return event, nil
+}
+
+func NewBedMovedToAnotherRoomEvent(roomId string) (hwes.Event, error) {
+	eventData := BedMovedToAnotherRoomEvent{
+		RoomId: roomId,
+	}
+	event := hwes.NewBaseEvent(BedUpdated)
```
PaulKalho, line 68: `BedUpdated` should be `BedMovedToAnotherRoom`

services/task-svc/internal/bed/bed.go
```diff
@@ -211,6 +214,12 @@ func (ServiceServer) UpdateBed(ctx context.Context, req *pb.UpdateBedRequest) (*
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	_ = events.DispatchBedUpdatedEvent(ctx, bedID, req.GetName())
+
+	if roomId.Valid {
+		_ = events.DispatchBedMovedToAnotherRoomEvent(ctx, bedID, roomId.UUID.String())
+	}
```
PaulKalho, line 221: Does it make sense to always dispatch the `BedMovedToAnotherRoomEvent` if the `roomId` is given & valid?

This means the `BedMovedToAnotherRoomEvent` will be dispatched even if it hasn't really changed but is inside the request, right?

PaulKalho (changes requested):
Erverything works now but I'm not too sure about the `BedMovedToAnotherRoom` Event.

services/task-svc/internal/bed/bed.go
MaxSchaefer: On the technical side: same goes for the `.UpdateBed`. We never check for diffs but just applying the requested changes by the user.
On the business logic side: There are two sides to this i guess:
- We can dispatch the event only if the value really changes
- We can dispatch the event every time the user dispatched the change
In event sourcing, both cases should result in the same state of our aggregate if our event handlers are idempotent. If the event gets consumed by services that are outside of this bounded context (task-svc) via an integration event (svc-to-svc events) the receiver can be sure that the bed actually moved.

I guess its a design decision? We should discuss this in the actual event sourcing implementation at https://github.com/helpwave/services/pull/587. I would close this conversation :)

PaulKalho (approved):

PR #585, opened 2023-12-16
sqlc: user and organization

### LINKED ISSUE #573: move user-svc's `user` models to SQLc

PaulKalho:
## Which issues does this pull request close
Closes #573, #574

## Important
- **AvatarUrl** made optional in GetOrganizationResponse
- **AvatarUrl** made optional in GetOrganizationsByUserResponse
- **AvatarUrl** made optional in GetOrganizationsForUserResponse
- **AvatarUrl** made optional in GetInvitationsByUserRespoonse

### Tests
Unfortunately I also had to edit some tests because I encountered an error which I didn't while writing the tests 

Affected: 
- GetInvitationsByOrganizations
- AddRemoveMember (bec. is using GetMembersByOrganization in test)
- GetMembersByOrganization 

=> All these Endpoints used the function `IsInOrganization` which was implemented wrong. It always returned true.

- RevokeInvitation

=> I'm not quite sure what gorm did there but I always got a wrong row in the query.


services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,122 @@
...
+	FROM memberships
+	WHERE user_id = @userID AND organization_id = @organizationID
+);
+
+-- name: DoesInvitationExist :one
+SELECT EXISTS (
+	SELECT 1
+	FROM invitations
+	WHERE (invitations.email = @email AND invitations.organization_id = @organization_id)
+		AND state = ANY(ARRAY[@state::int[]])
```
FoseFx, line 99:
```suggestion
		AND state = ANY(ARRAY[@states::int[]])
```

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,122 @@
...
+	SELECT 1
+	FROM memberships
+	WHERE user_id = @userID AND organization_id = @organizationID
+);
+
+-- name: DoesInvitationExist :one
+SELECT EXISTS (
+	SELECT 1
+	FROM invitations
+	WHERE (invitations.email = @email AND invitations.organization_id = @organization_id)
```
FoseFx, line 98: nit: `invitations.` is not needed here

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,122 @@
...
+);
+
+-- name: GetInvitationsWithOrganizationByUser :many
+SELECT
+	sqlc.embed(organizations),
+	invitations.*
+FROM invitations
+	JOIN organizations ON invitations.organization_id = organizations.id
+WHERE email = @email
+	AND state = coalesce(sqlc.narg('state'), state);
```
FoseFx, line 109: an OR ... IS NULL might be faster here

Without knowing about what's going on internally, I believe this has to pull out the state of every row and compare it to itself, if state is NULL

FoseFx (changes requested):
Nice work! Some of the tests need to be looked into though.

PaulKalho:
@FoseFx thanks for the Review. 
I've cleaned up the sql statements and refactored the tests. Left the `not a member` tests inside, but added the missing ones.

services/user-svc/internal/organization/organization.go
```diff
@@ -685,88 +701,66 @@ func (s ServiceServer) RevokeInvitation(ctx context.Context, req *pb.RevokeInvit
...
+		LongName:        attr.LongName,
+		ShortName:       attr.ShortName,
+		ContactEmail:    attr.ContactEmail,
+		AvatarUrl:       attr.AvatarUrl,
+		IsPersonal:      attr.IsPersonal,
+		CreatedByUserID: userID,
 	})
-
 	if err != nil {
 		return nil, err
```
FoseFx, line 723: I believe you have to rollback the transaction before we return. From what I read in the docs you could simply add a `defer tx.Rollback()` after creating the transaction, which will cause the rollback to be attempted before any return. If we commit before, the rollback won't happen. But please read up on that again :D

services/user-svc/internal/user/user.go
```diff
@@ -31,86 +34,93 @@ func NewServiceServer() *ServiceServer {
...
+	} else if user == nil {
+		return nil, status.Error(codes.Internal, "record not found")
 	}
 
 	return &pb.ReadPublicProfileResponse{
 		Id:        user.ID.String(),
 		Name:      user.Name,
 		Nickname:  user.Nickname,
-		AvatarUrl: user.Avatar,
+		AvatarUrl: *user.AvatarUrl,
```
FoseFx, line 98: AvatarUrl is not `NOT NULL`.
THIS WILL CAUSE NIL-DEREFERENCES!

FoseFx (changes requested):
Mostly small things, but the transaction rollback and NIL-deref bugs *need* to be fixed

services/user-svc/internal/user/user.go
PaulKalho: Hm true, thought because of the default value this shouldn't be a problem.
Forgot that it can theoretically be NULL if this was inserted intentionally, will fix that.

services/user-svc/internal/organization/organization.go
```diff
@@ -326,9 +344,9 @@ func (s ServiceServer) RemoveMember(ctx context.Context, req *pb.RemoveMemberReq
 }
 
 func (s ServiceServer) InviteMember(ctx context.Context, req *pb.InviteMemberRequest) (*pb.InviteMemberResponse, error) {
```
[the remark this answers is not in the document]
PaulKalho: that's a good call, ty

services/user-svc/internal/organization/organization.go
PaulKalho: Yes you are 100% right, must have missed that.

MaxSchaefer:
@FoseFx

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,139 @@
...
+-- name: GetInvitationConditions :one
+SELECT
+	EXISTS (
+		SELECT 1
+		FROM memberships
+				 JOIN users ON users.id = memberships.user_id
+		WHERE memberships.organization_id = @organization_id
+		  AND users.email = @email
+		LIMIT 1
+	) AS isInOrganitationByEmail,
```
FoseFx, line 103:
```suggestion
	) AS is_in_organization_by_email,
```

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,139 @@
...
+
+	EXISTS (
+		SELECT 1
+		FROM organizations
+		WHERE (id = @organization_id)
+	) AS doesOrganizationExist,
```
FoseFx, line 109:
```suggestion
	) AS does_organization_exist,
```

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,139 @@
...
+
+	EXISTS (
+		SELECT 1
+		FROM invitations
+		WHERE (email = @email AND organization_id = @organization_id)
+		  AND state = ANY(ARRAY[@states::int[]])
+	) AS DoesInvitationExist;
```
FoseFx, line 116:
```suggestion
	) AS does_invitation_exist;
```

services/user-svc/repos/organization_repo.sql
```diff
@@ -0,0 +1,139 @@
...
+		SELECT 1
+		FROM organizations
+		WHERE (id = @organization_id)
+	) AS doesOrganizationExist,
+
+	EXISTS (
+		SELECT 1
+		FROM invitations
+		WHERE (email = @email AND organization_id = @organization_id)
+		  AND state = ANY(ARRAY[@states::int[]])
```
FoseFx, line 115: what does the `ARRAY[..]` do / why is it needed?

services/user-svc/internal/organization/organization.go
```diff
@@ -709,6 +692,10 @@ func CreateOrganizationAndAddUser(ctx context.Context, attr organization_repo.Or
 		return nil, err
 	}
 
+	defer func() {
+		_ = tx.Rollback(ctx)
```
FoseFx, line 696:
```suggestion
		// TODO: log Rollback errors
		_ = tx.Rollback(ctx)
```

FoseFx (approved):
Nice! Only a few finishing touches, then you can merge!

services/user-svc/repos/organization_repo.sql
PaulKalho: You're right, we dont't need it at all. Wonder why I put it there originally...

PR #587, opened 2023-12-20
feat: kick-off event sourced tasks-svc

### LINKED ISSUE #344: Rename `task-svc` to `tasks-svc`
### Describe the chore

The current `task-svc`service contains all relevant entities and business logic for the `helpwave tasks` product. We are slicing by products and cross-cutting concerns.

MaxSchaefer:
This pull request starts our journey towards event sourcing and domain-driven-design in general. Therefore I've implemented `libs/hwes` that gets consumed by `tasks-svc` (and every other service that wants to utilizing event sourcing). `libs/hwes` introduces the concepts of Commands, Events, Aggregates and Aggregates stores. These concepts are implemented by `tasks-svc`, for example `TaskAggregate` or `CreateTaskCommand` or `TaskCreatedEvent`.

We already did a meeting on this new architecture but here is an outline:
1. Request gets handled by command or query (cq)
2. CQ loads the aggregate on which behalf to act on (by ID of aggregate)
2.1 Create empty aggregate by the passed ID
2.2 Load all events from the event store for that aggregate (every aggregate has its own event stream)
2.3 Applying all events in order on the aggregate
3. CQ triggers business logic on aggregates that resolves in new uncommitted events
5. Before returning the request back to the caller of the CQ, persist all events in our event store.
4.1 `libs/hwes` calculates the expected revision of the targeted event stream by combining the applied events of step 2. and the uncommitted events of step 3. .
4.2.a If the revision cannot be reached due to concurrent changes to the targeted event stream the request fails.
4.2.b If the revision can be reached during persistence, the events got stored successfully.

Step 4.2 describes [optimistic concurrency](https://developers.eventstore.com/clients/grpc/appending-events.html#handling-concurrency). The application needs to keep track of versioning. We could apply all events, regarding there version but this can result in [data conflicts and issues with idempotency](https://developers.eventstore.com/clients/dotnet/21.2/appending.html#idempotence). Its quite unlikely that Step 4.2.a occur but we need to keep track of this architecture trade-off with future metrics and tracing.

`tasks-svc` is not yet production ready. We can either integrate the event sourced task domain into the current `task-svc` or the other way around. Upcoming issues will work towards the production readiness.

To-dos for later issues
- Snapshots
- Projections and Subscriptions
- Identity context in events

Also closes #344


services/tasks-svc/internal/task/commands/v1/update_task.go
```diff
@@ -0,0 +1,51 @@
...
+	if command.Name != nil {
+		if err := a.UpdateName(ctx, a.Task.Name, *command.Name); err != nil {
+			return err
+		}
+	}
+
+	if command.Description != nil {
+		if err := a.UpdateDescription(ctx, a.Task.Description, *command.Description); err != nil {
+			return err
+		}
```
PaulKalho, line 47: Because the attribute `description` is missing on the `CreateTaskCommand` `a.Task.Description` will always be empty here on the first `TaskDescriptionUpdatedEvent` even if the description has been set initially.

PaulKalho (changes requested):
All in all that looks good. Didn't find any mistakes in the `aggregate` / `aggregate_store`.
Anyways the missing `description` in the `TaskCreated`-Event has to be fixed.

MaxSchaefer:
Discussion!
https://github.com/helpwave/services/pull/582#discussion_r1437002287

services/tasks-svc/internal/task/api/grpc.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	command := commandsV1.NewAssignTaskCommand(taskID, userID)
+	if err := s.service.Commands.AssignTask.Handle(ctx, command); err != nil {
+		return nil, err
+	}
+
+	return &pb.AssignTaskResponse{}, nil
+}
+
```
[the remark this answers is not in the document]
MaxSchaefer: I've added a comment.

services/tasks-svc/internal/task/api/grpc.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	userID, err := uuid.Parse(req.UserId)
+	if err != nil {
+		return nil, err
+	}
+
+	command := commandsV1.NewAssignTaskCommand(taskID, userID)
```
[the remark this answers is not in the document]
MaxSchaefer: I've added a comment.

proto/services/tasks_svc/v1/task_svc.proto
```diff
@@ -0,0 +1,161 @@
...
+//
+// UpdateTask
+//
+
+message UpdateTaskRequest {
+  string id = 1; // @gotags: validate:"uuid4"
+  optional string name = 2;
+  optional string description = 3;
+  optional google.protobuf.Timestamp due_at = 4;
+}
```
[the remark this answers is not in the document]
MaxSchaefer: A seperate endpoint should be more explicit.

proto/services/tasks_svc/v1/task_svc.proto
```diff
@@ -0,0 +1,161 @@
...
+//
+
+message UpdateSubtaskRequest {
+  string task_id = 1; // @gotags: validate:"uuid4"
+  string subtask_id = 2; // @gotags: validate:"uuid4"
+  Subtask subtask = 3;
+
+  message Subtask {
+    optional string name = 1;
+  }
```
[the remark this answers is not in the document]
MaxSchaefer: If we move the `subtask_id` into a sub-message. This sub-message can be nillable in Go. Therefore we need to do  an additional nillable-check like the following example. If we leave the `subtask_id` out of the nillable sub-message, this gets handled by grpc and our validation. What do you think @PaulKalho?

```
if req.Subtask == nil {
	return &pb.UpdateSubtaskResponse{}, nil
}

subtaskID, err := uuid.Parse(req.Subtask.GetSubtaskId())
if err != nil {
	return nil, err
}
```

services/tasks-svc/internal/task/commands/v1/update_task.go
MaxSchaefer: The description is an empty string `""` when the description gets not specified during creation. See `models.NewTask()`.

proto/services/tasks_svc/v1/task_svc.proto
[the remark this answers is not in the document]
PaulKalho: Hmm true, then we could probably just leave it how it is. Does not make any major difference in my opinion, do what you think is better :)

PaulKalho (approved):
Changes look good in my opinion.

Before merging we should probably still discuss https://github.com/helpwave/services/pull/587#issuecomment-1878984959

PaulKalho:
> Discussion! [#582 (comment)](https://github.com/helpwave/services/pull/582#discussion_r1437002287)

Imo we should only emit these kind of events if the value really changes. We will most probably not get around to implement some kind of logic to compare old and new values (we need that for the `taskSelfAssigned` Event anyways) and it will ensure tidiness in our eventstore.

But as @MaxSchaefer already said, the state of our aggregate will be the same no matter for which option we decide.

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,184 @@
...
+func (a *AggregateBase) RaiseEvent(event Event) error {
+	if event.GetAggregateID() != a.GetID() {
+		return errors.New("invalid aggregate for event")
+	}
+
+	if event.GetVersion() < a.GetVersion() {
+		return errors.New("invalid version")
+	}
+
+	event.SetAggregateType(a.GetType()) // TODO: Is this really necessary?
```
FoseFx, line 175: see above

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,184 @@
...
+		return err
+	}
+
+	a.version++
+	event.SetVersion(a.GetVersion())
+	a.uncommittedEvents = append(a.uncommittedEvents, event)
+	return nil
+}
+
+// RaiseEvent should be called after all events are loaded though an aggregate store.
```
FoseFx, line 164: Please first state what the function does, before explaining when / how to use it.
When should I use Apply instead? We should also change the name. Mabye rename Apply to ApplyUncomitted and this one to ApplyComitted or sth.

libs/hwes/aggregate_store.go
```diff
@@ -0,0 +1,16 @@
+package hwes
+
+import (
+	"context"
+)
+
+type AggregateStore interface {
```
FoseFx, line 7: explain what an AggregateStore is

libs/hwes/command.go
```diff
@@ -0,0 +1,19 @@
+package hwes
+
+import "github.com/google/uuid"
+
+type Command interface {
```
FoseFx, line 5: explain what a command is and how to use it

libs/hwes/event.go
```diff
@@ -4,27 +4,95 @@ import (
...
-	Type      string
-	Data      []byte
-	Timestamp time.Time
+	EventID       uuid.UUID
+	EventType     string
+	AggregateID   uuid.UUID
+	AggregateType AggregateType
+	Data          []byte
+	Timestamp     time.Time
+	Version       int64 // TODO: uint64?
```
FoseFx, line 18: We should clear this up!

FoseFx (commented):
Nowhere near done reviewing this. This review is extremely unpolished, sorry , but I want to give you early feedback already.

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,184 @@
+package hwes
+
+import (
+	"errors"
+	"fmt"
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+)
+
+// When is a per aggregate defined function that
```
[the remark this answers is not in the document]
MaxSchaefer: Lets ditch this `When`-Interface! The event handlers was getting registered Inside the concrete implementation of `When`. I've added `.RegisterEventHandler()` instead.

libs/hwes/command.go
```diff
@@ -0,0 +1,29 @@
+package hwes
+
+import "github.com/google/uuid"
+
+// CommandBase acts as a container for the targeted AggregateID
+// A concrete command in our CQRS pattern must call NewCommandBase
```
FoseFx, line 6: > must 

no, it doesn't

if we want to enforce anything, there should be an interface for it

libs/hwes/event.go
```diff
@@ -2,29 +2,124 @@ package hwes
...
+	EventType     string
+	AggregateID   uuid.UUID
+	AggregateType AggregateType
+	Data          []byte
+	Timestamp     time.Time
+	Version       uint64
 }
 
-func NewBaseEvent(eventType string) Event {
+func NewEventBase(aggregate Aggregate, eventType string) Event {
```
FoseFx, line 22: this is just a constructor for `Event`, there is no `EventBase`, nor should there be one, as an event is never wrapped, as data is set in the event object

libs/hwes/event.go
```diff
@@ -2,29 +2,124 @@ package hwes
...
-		Timestamp: time.Now().UTC(),
+		EventID:       uuid.New(),
+		EventType:     eventType,
+		AggregateID:   aggregate.GetID(),
+		AggregateType: aggregate.GetType(),
+		Timestamp:     time.Now().UTC(),
 	}
 }
 
+func NewEventBaseWithPayload(aggregate Aggregate, eventType string, payload interface{}) (Event, error) {
```
FoseFx, line 32:
```suggestion
func NewEventBaseWithData(aggregate Aggregate, eventType string, data interface{}) (Event, error) {
```

FoseFx, line 32: mention that the payload will be json marshalled. that's important information for a user of this function

maybe just mention `SetJsonData`

libs/hwes/event.go
```diff
@@ -2,29 +2,124 @@ package hwes
...
+	eventBase := NewEventBase(aggregate, eventType)
+	if err := eventBase.SetJsonData(&payload); err != nil {
+		return Event{}, err
+	}
+	return eventBase, nil
+}
+
+// ResolveAggregateIDAndTypeFromStreamID converts a streamID to an aggregateType and aggregateID
```
FoseFx, line 40:
```suggestion
// ResolveAggregateIDAndTypeFromStreamID extracts the aggregateType and aggregateID of a given streamID
```

libs/hwes/event.go
```diff
@@ -2,29 +2,124 @@ package hwes
...
+// See aggregate.GetStreamID
+//
+// # Example
+//
+// StreamID:		task-d9027be3-d00f-4eec-b50e-5f489df20433
+// AggregateType: 	task
+// AggregateID: 	d9027be3-d00f-4eec-b50e-5f489df20433
+func ResolveAggregateIDAndTypeFromStreamID(streamID string) (AggregateType, uuid.UUID, error) {
```
FoseFx, line 48: nit:

using a syntax like this:
```suggestion
func ResolveAggregateIDAndTypeFromStreamID(streamID string) (aggregateType AggregateType, aggregateId uuid.UUID, err error) {
```

and setting the returned variable names makes the error path's returns easier to read

libs/hwes/event.go
```diff
@@ -2,29 +2,124 @@ package hwes
...
+	if err != nil {
+		return "", uuid.UUID{}, err
+	}
+
+	return aggregateType, aggregateID, err
+}
+
+// NewEventFromRecordedEvent is a helper function for EventStore.
+// This function transforms esdb.RecordedEvent to hwes.Event.
+// We expect that the StreamID of the aggregate is in the format of "[aggregateType]-[aggregateID]".
```
FoseFx, line 75: is there a reason to let the user of this library know about stream ids at all? can't we just build/parse it when needed?

libs/hwes/eventstoredb/aggregate_store.go
```diff
@@ -0,0 +1,158 @@
...
+	"hwutil"
+	"io"
+	"math"
+)
+
+type AggregateStore struct {
+	es *esdb.Client
+}
+
+type getExpectedRevision = func(ctx context.Context, aggregate hwes.Aggregate) (esdb.ExpectedRevision, error)
```
FoseFx, line 17: explain the point of this type

libs/hwes/eventstoredb/aggregate_store.go
```diff
@@ -0,0 +1,158 @@
...
+
+func NewAggregateStore(es *esdb.Client) *AggregateStore {
+	return &AggregateStore{es: es}
+}
+
```
FoseFx, line 22: nit: mention which interface is implemented here

libs/hwes/eventstoredb/aggregate_store.go
```diff
@@ -0,0 +1,158 @@
...
+func (a *AggregateStore) Load(ctx context.Context, aggregate hwes.Aggregate) error {
+	stream, err := a.es.ReadStream(ctx, aggregate.GetStreamID(), esdb.ReadStreamOptions{}, math.MaxUint64) // MaxUint64 for "all" events
+	if err != nil {
+		return err
+	}
+	defer stream.Close()
```
FoseFx, line 28: move this in front of the error handling for good measure

services/task-svc/internal/bed/bed.go
```diff
@@ -55,8 +54,6 @@ func (ServiceServer) CreateBed(ctx context.Context, req *pb.CreateBedRequest) (*
 		Str("name", bed.Name).
 		Msg("bed created")
 
-	_ = events.DispatchBedCreatedEvent(ctx, bed.ID, bed.Name)
```
FoseFx, line 58: why?

services/task-svc/internal/task/task.go
```diff
@@ -579,15 +578,6 @@ func (ServiceServer) AssignTaskToUser(ctx context.Context, req *pb.AssignTaskToU
 		Str("userID", userId.String()).
 		Msg("user assigned")
 
-	if currentUserId, err := common.GetUserID(ctx); err == nil {
```
FoseFx, line 582: also why?

services/tasks-svc/main.go
```diff
@@ -0,0 +1,29 @@
...
+const ServiceName = "tasks-svc"
+
+// Version is set at compile time
+var Version string
+
+func main() {
+	common.Setup(ServiceName, Version, true)
+
+	eventStore := eventstoredb.SetupEventStoreByEnv()
+	aggregateStore := eventstoredb.NewAggregateStore(eventStore)
```
FoseFx, line 21: I think there should be only one line for this. If I only need the eventStore to create the aggregateStore, why not just create the aggregateStore?

FoseFx (commented):
still not close to reviewing everything

libs/hwes/eventstoredb/setup.go
```diff
@@ -0,0 +1,33 @@
...
+
+func SetupEventStoreByEnv() *esdb.Client {
+	connectionString := hwutil.MustGetEnv("EVENTSTORE_CS")
+	return SetupEventStore(connectionString)
+}
+
+func SetupEventStore(connectionString string) *esdb.Client {
+	config, err := esdb.ParseConnectionString(connectionString)
+	if err != nil {
+		log.Fatal().Err(err).Msg("invalid eventstore connection string")
```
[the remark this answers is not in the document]
MaxSchaefer: connection string contains sensitive data like the eventstoredb password

services/task-svc/internal/task/task.go
MaxSchaefer: The initial usage of `hwes` in `task-svc` was just testing, not for event sourcing purposes.

services/task-svc/internal/bed/bed.go
MaxSchaefer: The initial usage of `hwes` in `task-svc` was just testing, not for event sourcing purposes.

services/tasks-svc/main.go
MaxSchaefer: Later on we need to share the eventStore instance for projections.

libs/hwes/eventstoredb/aggregate_store.go
MaxSchaefer: Oh yes! Thats quite important!

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,196 @@
...
+func (a *AggregateBase) GetVersion() uint64 {
+	return a.version
+}
+
+// RegisterEventListener registers the callback function for a specific event type
+// If you call RegisterEventListener multiple times for the same eventType, the last eventHandler gets registered.
+func (a *AggregateBase) RegisterEventListener(eventType string, eventHandler eventHandler) *AggregateBase {
+	_, found := a.eventHandlers[eventType]
+	if found {
+		log.Warn().
```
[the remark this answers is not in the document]
MaxSchaefer: @FoseFx yeah true. `Fatal` would be to hard. `Error` is okay I guess as a runtime error. I would not like to have an error returned due to the beautiful daisy chaining of of `. RegisterEventListener()` ... 😅

services/tasks-svc/internal/task/api/grpc.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (s *TaskGrpcService) UpdateTask(ctx context.Context, req *pb.UpdateTaskRequest) (*pb.UpdateTaskResponse, error) {
+	taskID, err := uuid.Parse(req.GetId())
+	if err != nil {
+		return nil, err
+	}
+
+	command := commandsV1.NewUpdateTaskCommand(taskID, req.Name, req.Description)
+	if err := s.service.Commands.UpdateTask.Handle(ctx, command); err != nil {
```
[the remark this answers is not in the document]
MaxSchaefer: DONE!

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,196 @@
...
+
+type AggregateType string
+
+// Aggregate reflects a concept of Domain-Driven-Design.
+// This interface describes the contracts that are used
+// by an AggregateStore to persists this aggregate in an
+// event-sourcing manner
+type Aggregate interface {
+	GetID() uuid.UUID
+	GetStreamID() string // TODO: Make private https://github.com/helpwave/services/pull/587#discussion_r1451788115
```
FoseFx, line 19: this TODO is still open

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,196 @@
...
+	atype   AggregateType
+	version uint64
+
+	eventHandlers map[string]eventHandler
+
+	appliedEvents     []Event
+	uncommittedEvents []Event
+}
+
+// NewAggregateBase must be called by a concrete aggregate during its initialization
```
FoseFx, line 45:
```suggestion
// NewAggregateBase constructs a new AggregateBase.
// When implementing a new, concrete, aggregate struct you can embed an AggregateBase into it.
// This way you get an implementation of the Aggregate interface for free.
```

libs/hwes/aggregate.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+type eventHandler func(evt Event) error
+
+type AggregateType string
+
+// Aggregate reflects a concept of Domain-Driven-Design.
+// This interface describes the contracts that are used
+// by an AggregateStore to persists this aggregate in an
+// event-sourcing manner
```
FoseFx, line 16:
```suggestion
// event-sourcing manner
// Also see AggregateBase and NewAggregateBase, you probably don't want to implement this interface by hand
```

libs/hwes/event.go
```diff
@@ -2,29 +2,128 @@ package hwes
...
+	aggregateType = AggregateType(aggregateTypeStr)
+	if aggregateType == "" {
+		err = fmt.Errorf("resolved empty aggregateType from streamID '%s'", streamID)
+		return
+	}
+
+	aggregateID, err = uuid.Parse(aggregateIDStr)
+	if err != nil {
+		return
+	}
```
FoseFx, line 72:
```suggestion
```

libs/hwes/eventstoredb/setup.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+	"github.com/EventStore/EventStore-Client-Go/esdb"
+	"github.com/rs/zerolog/log"
+	"hwutil"
+)
+
+func SetupEventStoreByEnv() *esdb.Client {
+	connectionString := hwutil.MustGetEnv("EVENTSTORE_CS")
```
FoseFx, line 11: is this (and other) env variable(s) documented in the root's README?

services/tasks-svc/internal/task/aggregate/actions.go
```diff
@@ -0,0 +1,107 @@
+package aggregate
+
+import (
+	"context"
+	pb "gen/proto/services/tasks_svc/v1"
+	"github.com/google/uuid"
+	eventsV1 "tasks-svc/internal/task/events/v1"
```
FoseFx, line 7: pretty sure this name won't survive for long, how about `taskEvents`?

services/tasks-svc/internal/task/aggregate/aggregate.go
```diff
@@ -0,0 +1,251 @@
...
+
+func LoadTaskAggregate(ctx context.Context, as hwes.AggregateStore, id uuid.UUID) (*TaskAggregate, error) {
+	task := NewTaskAggregate(id)
+	if err := as.Load(ctx, task); err != nil {
+		return nil, err
+	}
+	return task, nil
+}
+
+func (a *TaskAggregate) initEventListener() {
```
FoseFx, line 37:
```suggestion
func (a *TaskAggregate) initEventListeners() {
```

services/tasks-svc/internal/task/aggregate/aggregate.go
```diff
@@ -0,0 +1,251 @@
...
+type TaskAggregate struct {
+	*hwes.AggregateBase
+	Task *models.Task
+}
+
+func NewTaskAggregate(id uuid.UUID) *TaskAggregate {
+	aggregate := &TaskAggregate{Task: models.NewTask()}
+	aggregate.AggregateBase = hwes.NewAggregateBase(TaskAggregateType, id)
+	aggregate.Task.ID = id
+	aggregate.initEventListener()
```
MaxSchaefer, line 25:
```suggestion
	aggregate.initEventListeners()
```

libs/hwes/aggregate.go
MaxSchaefer: Done. I reframed the function to respect separation of concerns.

FoseFx: Maybe GitHub shows me an old version (can't really check rn, am on mobile), but this comment still exists?

FoseFx (approved):
🥂

PR #592, opened 2023-12-29
ci: lint and test go libs

### LINKED ISSUE #40: libs not linted in CI
### Describe the bug
We only lint each service, but never the libs.

MaxSchaefer:
Until now, `libs/` was not linted and tested in our CI.
- Most of the applied changes to satisfy our linter was fixed by running `go mod tidy`.
- `libs/common` got https://github.com/helpwave/services/pull/592/files#diff-1a716d5bd39b0139f3dbdd6164329cbe734a534fd569860ccc722ea14a46c1a1L37 removed. [Manually seeding `math/rand` is no longer needed since Go 1.20](https://tip.golang.org/doc/go1.20)
- Exception for `libs/proto_helpers` due to the fact that this package does not require any direct external packages. If the CI runs `go get .`, the command exits (1) with `go: .: no package to get in current directory`.

Closes #40


.github/workflows/cicd.yaml
```diff
@@ -55,6 +56,7 @@ jobs:
         run: |
           if [[ "$libsChanged" == "true" ]]; then
             echo "services=$(ls services | grep -v 'impulse_svc' | uniq | jq -R . | jq -sc .)" >> $GITHUB_OUTPUT
+            echo "libs=$(ls libs | grep -v 'proto_helpers' | uniq | jq -R . | jq -sc .)" >> $GITHUB_OUTPUT
```
FoseFx, line 59: just add a comment explaining this line, and why proto_helpers should be excluded

FoseFx (approved):

PR #601, opened 2024-01-09
Tracing pt.1

### LINKED ISSUE #596: Tracing
### Describe the feature

We need to decide on a tracing provider. The two common providers are OpenZipkin and Jaeger. Jaeger is part of the CNCF and written in Golang. We go with Jaeger.

To implement tracing, Jaeger utilizes [OpenTelemetry](https://opentelemetry.io/). Due to the setup of OpenTelemetry and observability in general we can either abstract this away in `libs/common` or creating a new lib like `libs/hwtel`.

The goal of this issue is to get OpenTelemetry (configured for Jaeger) into our repository. All of this should be behind a feature flag until the infrastructure deployment for Jaeger is done.

- [x] Add tracing of [OpenTelemetry](https://opentelemetry.io/)(for Jager) to `helpwave/services` as part of `libs/common` or a new lib `libs/hwtel` (depending on how extensive the setup is)
- [x] Add Jaeger deployment for development to `dev/docker-compose.yaml`
- [x] Trace some parts of `task-svc` and `libs/common` (middleware stuff for example)
- [x] Provide feature flag via environment variables for observability.

Observability gets important especially with a more complicated architecture as introduces in https://github.com/helpwave/services/pull/587, we need to make sure to provide such tools for monitoring and debugging.

### Additional context

This issue has the focus on tracing. OpenTelemetry also provides to ability to collect metrics. Should we use OpenTelemetry for metrics or do we want to go with the standard prometheus client?

FoseFx:
closes #596 

image

**Before merging we need to set `OTEL_DISABLE=true` for all staging services, or they will die on startup.**

- adds otel tracing setup
  supports zipkin, otlp (e.g. jaeger, elastic, ...) and oltp_http
- adds spans for various functions including interceptors, database connection and query
- registers SIGINT trap and cleans up server, otel and db
- jaeger added to docker compose

Currently, logs are not associated with spans yet, and I'm unsure about propagation.


MaxSchaefer:
> Why zipkin and not jaeger? Because it didn't work. After I started debugging with wireshark I realized it was not worth my time.

What problem did you have with Jaeger? I can take a quick look.

FoseFx:
It closed the tcp connection after the client sent the h2 header.

MaxSchaefer:
The connection fails on my machine between our services and Jaeger due to non existing SSL. We need the option for [WithInsecure](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.21.0#WithInsecure).

FoseFx:
No, [as documented](https://github.com/helpwave/services/pull/601/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R201) you can use the [environment](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.21.0) to configure the otel exporter.

For a local, insecure, connection you need to start the service with `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` and `OTEL_EXPORTER_OTLP_INSECURE=true`

MaxSchaefer (approved):

libs/common/grpc.go
```diff
@@ -66,14 +73,33 @@ func StartNewGRPCServer(addr string, registerServerHook func(*daprd.Server)) {
 		zlog.Warn().Msg("grpc reflection enabled")
 	}
 
-	zlog.Info().Str("addr", addr).Msg("starting grpc service")
+	interrupted, err := hwutil.RunUntilInterrupted(context.Background(), func() error {
```
MaxSchaefer, line 76: fyi: Later on we need to pass a caller-defined context here.
https://github.com/helpwave/services/blob/issue/x-spicedb/services/tasks-svc/main.go#L36

dapr_run.yaml
```diff
@@ -17,6 +17,7 @@ common:
     REDIS_ADDR: localhost:6379
     INSECURE_REDIS_NO_TLS: true
     ORGANIZATION_ID: 3b25c6f5-4705-4074-9fc6-a50c28eba405
+    OTEL_DISABLE: true
```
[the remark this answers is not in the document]
FoseFx: devenv up will not start jaeger, so this will, at best, lead to errors

libs/common/grpc.go
FoseFx: in pt.2 I already got rid of (most) global logger usage so this would be a [one line change](https://github.com/helpwave/services/pull/606/files#diff-c9f8f86c47f179c3476df0fdb02d379b26c1b903fcb445992bef82e9a23ce3baR44) ig

Issue #602, opened 2024-01-09
Service to Service trace propagation


FoseFx:
### Describe the feature

Ensure service to service trace propagation works. We currently only have a few cases where we do invoke other service's endpoints.

### Why is this feature needed?

asd

FoseFx:
Blocked by #618

PR #606, opened 2024-01-11
Tracing pt.2

### LINKED ISSUE #599: Add Logs to Traces
### Describe the feature

Follow up to #596

Add a hook to zerolog, like **[this](https://github.com/komuw/otero/blob/v0.0.1/log/zerolog.go).** Maybe [this](https://github.com/agoda-com/opentelemetry-logs-go) is interesting.

Needed for visibility into the system

### Additional context

Open Telemetry has three concerns: Traces, Metrics and Logs. This issue relates to the latter.

FoseFx:
closes #599

image

Stacked on #601

## Content

- adds [`zerologTraceHook`](https://github.com/helpwave/services/pull/606/files#diff-107df3cc945fe5b55204cc1308c80c4b788f24b0ac64064c9b169abbf99a815fR56) and [registers it with zerolog](https://github.com/helpwave/services/pull/606/files#diff-1321fd54337f6dee9fe48be93b7731b689a7ae52209c26052c6caf89d0c6cc7eR24)
- moves `StartSpan` and `StopSpanFromCtx` to `logging`
- `StartSpan` now injects a new logger, and context with said logger injected
- reduces usage of global `zlog.Logger` in favor of `.Ctx` loggers, which may be span-aware or can easily be made that
- [remove `span` log-field in logger middleware](https://github.com/helpwave/services/pull/606/files#diff-c9f8f86c47f179c3476df0fdb02d379b26c1b903fcb445992bef82e9a23ce3baL306), this is now handled by the otel propagator and then trace hook
- if and how the propagator works for [events](https://github.com/helpwave/services/pull/606/files#diff-c9f8f86c47f179c3476df0fdb02d379b26c1b903fcb445992bef82e9a23ce3baR328) is something for #602 to clear up
- some dependency cleaning up

Final note: This is using [events](https://opentelemetry.io/docs/instrumentation/go/manual/#events), as the [logs API is unstable](https://opentelemetry.io/docs/instrumentation/go/manual/#logs) in the go sdk.
There is another project that tries to imitate it, but it does not seem to be popular.
Events and LogRecords are the same and only differ in an additional `name` field for events.
The problem is the sdk creates events like this: `e := Event{Name: name, Attributes: c.Attributes(), Time: c.Timestamp()}`,
which means we have no way of setting typical [log records fields](https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition) like `SeverityText`. We, for now, have to shove everything into the `Attributes` field.
I don't think this is a problem, we just have to write different queries when looking for logs in the tracing backend.

## Log processing:

image


FoseFx:
btw with sufficient infrastructure in place we can deduplicate this and let the otel collector do log correlation, so we don't send logs to the tracing backend directly.

MaxSchaefer (approved):
Very nice. If this gets merged, we have strong tracing in place. Good job. The only thing I'm not quite happy about is that telemetry is part of the `logging` package. The syntax `logging.StartSpan ...` is not _so_ satisfying.

FoseFx:
Rename the logging package to telemetry?

MaxSchaefer:
> Rename the logging package to telemetry?

My comment was not constructive. Thanks for taking care of that 😅 Yes, `telemetry` sounds fine. The exported logging-only functions are quite expressive and therefore the broader term `telemetry` should be fine.

PR #607, opened 2024-01-12
buf: remove compiler version comments

FoseFx:
Everytime buf updates their compiler our CI breaks, as there is a diff.
This is a somewhat cheap fix for this.

I already pushed the new ghcr.io/helpwave/service-preproc:edge


MaxSchaefer (approved):
@FoseFx can we pin the compiler version?

FoseFx:
I first thought no, because the python plugin was version pinned already (and the python compiler version changed recently, causing all this), but not every python plugin was pinned.

Let's try that first. Closing in favour of #608

PR #617, opened 2024-01-27
initialize property service

### LINKED ISSUE #616: Initialize Property-Service
The goal of this issue is to initialize and bootstrap the `propery-svc` with some hello-world aggregates, commands, queries and events.

PaulKalho:
closes #616


MaxSchaefer:
I've update my To-do comment on https://github.com/helpwave/helpwave/issues/109#issuecomment-1919712778

> - The soft-required feature is "view-only" and should be configured for example on the ward level
I guess thats clear.

> - If a data-type provides the ability to pre-define options, the user should still be able self-create and add custom ones during property assignment
The new property option should be added an option to the property with a "custom"-Flag on `true`.

services/property-svc/internal/property-set/aggregate/aggregate.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+	"github.com/google/uuid"
+	"hwes"
+	propertySetEventsV1 "property-svc/internal/property-set/events/v1"
+	"property-svc/internal/property-set/models"
+)
+
+const PropertySetAggregateType = "propertySet"
```
MaxSchaefer, line 11:
```suggestion
const PropertySetAggregateType = "property_set"
```

proto/services/property_svc/v1/property_value_svc.proto
```diff
@@ -0,0 +1,58 @@
...
+
+//
+// CreatePropertyValue
+//
+message CreatePropertyValueRequest{
+  string subject_id = 1;
+  string subject_type = 2;
+  string property_id = 3;
+  oneof value {
+    bool is_undefined = 4; // if it exists, it's true
```
MaxSchaefer, line 17: Where did we define that?

services/property-svc/internal/property/aggregate/aggregate.go
```diff
@@ -0,0 +1,59 @@
...
+	if err := evt.GetJsonData(&payload); err != nil {
+		return err
+	}
+
+	subjectID, err := uuid.Parse(payload.SubjectID)
+	if err != nil {
+		return err
+	}
+
+	fieldType := (pb.FieldType)(pb.FieldType_value[payload.FieldType])
```
MaxSchaefer, line 51: Is this safe? Tbd

services/property-svc/internal/property/models/property.go
```diff
@@ -0,0 +1,45 @@
...
+	SubjectID   uuid.UUID
+	SubjectType string
+	FieldType   pb.FieldType
+
+	Name           string
+	Description    string
+	IsArchived     bool
+	IsSoftRequired bool
+
+	PriorityDisplayedAtSubjectsIDs []string
```
MaxSchaefer, line 19:
```suggestion
```

MaxSchaefer (changes requested):

proto/services/property_svc/v1/property_value_svc.proto
MaxSchaefer: Will be addressed after https://github.com/helpwave/services/pull/619

MaxSchaefer (approved):
Very nice! The implemented logic gets eventually fully replaced by https://github.com/helpwave/services/pull/619

PR #618, opened 2024-01-31
Docker All The Way + Renovate Dependencies

### LINKED ISSUE #556: pubsub not running in devenv
### Describe the chore

Start dapr's pubsub with deven up

FoseFx:
- Adds the `dev-go` image
- Adds the `dev-go-custom` Dockerfile (supposed to be built on your machine), depends on `dev-go`
- Streamlines development into one Docker compose
- Adds `./up.sh`, `./down.sh` and `./shell.sh` to get into the dev environment quickly
- Adds devcontainer.json
- make tracing work for apisix and single services

---

- Pins dependencies in Dockerfiles and github workflows
- Let Renovate know about them, so they are managed in sync

---

Closes #556


FoseFx:
What versions of docker and compose and are you on?

PaulKalho:
> What versions of docker and compose and are you on?

I'm on docker version: 24.0.7
and compose version: 2.23.3

FoseFx:
https://github.com/helpwave/services/blob/main/libs/common/setup.go#L109

@MaxSchaefer is there a reason why we listen only on 127.0.0.1 when DAPR_PORT is set? This looks deliberate.

FoseFx:
Given that connections work when we listen on all interfaces, I assume that docker forwards from a different interface than the loop-back interface.

FoseFx:
Which I think is also true for sidecars in a k8s deployments

MaxSchaefer:
Its working quite well on GitHub Codespaces. I will also test it locally.
@PaulKalho is your issue fixed?

FoseFx:
Also, is the upstream discovery issue on Windows fixed now? @DasProffi

DasProffi:
> Also, is the upstream discovery issue on Windows fixed now? @DasProffi

I still get the no healthy service error in the browser console `"no healthy services found with AppID 'user-svc'"`

PaulKalho:
> @PaulKalho is your issue fixed?

@MaxSchaefer @FoseFx works like a charm. :)

FoseFx:
Approval?

MaxSchaefer:
> Approval?

My review and approval is on the way. I've pinged @Paul-B98 to test run this setup (no full review).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "dockerComposeFile": "../dev/docker-compose.yaml",
+  "service": "devcontainer",
+  "remoteUser": "helpwave",
+  "workspaceFolder": "/helpwave",
+  "overrideCommand": false
+}
```
Paul-B98, line 7: you could utilize the `postCreateCommand` to install dependencies like pre commit hooks or the `initializeCommand` to regularly pull docker images so that they don’t get outdated. 

It could also be helpful to use the `customizations` setting to add extensions or settings for vscode when used with a devcontainer.

dev/docker-compose.yaml
```diff
@@ -46,71 +43,111 @@ services:
...
+    restart: unless-stopped
+    network_mode: "service:devcontainer" # share ports with devcontainer
 
-  ory-svc:
+  # devcontainer is where the developer should do their development in
+  devcontainer:
     build:
-      context: ..
-      dockerfile: Dockerfile.standalone
+      context: ../images/dev-go-custom
```
Paul-B98, line 126: Could it be more beneficial to use an officially maintained devcontainer image for go that build your own?

* https://hub.docker.com/_/microsoft-devcontainers
* https://hub.docker.com/_/microsoft-vscode-devcontainers

.devcontainer/devcontainer.json
FoseFx: Great points, and I already did think about them!


> postCreateCommand

We have the entrypoint.sh for that. It runs when using docker (standalone), docker compose and using devcontainers. I want these environments to be as close to each other as possible.

> regularly pull docker images

By pinning the version of the image in git this is done automatically when starting the environment. This way it's also clear which versions of the image work with what states of the codebase. We can checkout old versions and still get a runnning system

dev/docker-compose.yaml
FoseFx: From what I can gather, these images are not meant to be run standalone.

MaxSchaefer (approved):
Very nice! Great work. Already v3 of our development environment :) 🎉

PR #625, opened 2024-02-06
SQLc: task templates

FoseFx:
## Which issues does this pull request close?
closes #623

GetAllTaskTemplatesByCreator changed in its semantic to reflect what the name says. Adds `private_only` for the frontend

Issue #569, opened 2023-12-02
Add Integration Tests


FoseFx:
At least for the to be refactored endpoints.

task_svc
- [ ] CreateWard
- [x] GetWard (#578)
- [x] GetWards (#578)
- [ ] GetRecentWards (#576)
- [ ] UpdateWard
- [ ] DeleteWard
- [x] GetWardOverviews (#578)
- [x] GetWardDetails (#578)
- [ ] CreateRoom
- [x] GetRoom (#567)
- [x] GetRooms (#567)
- [x] GetRoomsByWard (#567)
- [ ] UpdateRoom
- [ ] DeleteRoom
- [x] GetRoomOverviewsByWard (#567)
- [ ] CreatePatient
- [x] GetPatient (#567)
- [x] GetPatientByBed (#567)
- [x] GetPatientsByWard (#567)
- [x] GetPatientAssignmentByWard (#567)
- [x] GetRecentPatients (#567)
- [ ] UpdatePatient
- [x] AssignBed (#567)
- [x] UnassignBed (#567)
- [x] DischargePatient (#567)
- [x] GetPatientDetails (#567)
- [x] GetPatientList (#567)
- [x] DeletePatient (#567)
- [x] ReadmitPatient (#567)
- [x] CreateTaskTemplate (#630)
- [x] GetAllTaskTemplates (#630)
- [x] DeleteTaskTemplate (#630)
- [x] DeleteTaskTemplateSubTask (#630)
- [x] UpdateTaskTemplate (#630)
- [x] UpdateTaskTemplateSubTask (#630)
- [x] CreateTaskTemplateSubTask (#630)
- [x] GetAllTaskTemplatesByCreator (#630)
- [x] GetAllTaskTemplatesByWard (#630)
- [x] CreateTask (#593)
- [x] GetTask (#567)
- [x] GetTasksByPatient (#593)
- [x] GetTasksByPatientSortedByStatus (#593)
- [x] GetAssignedTasks (#593)
- [x] UpdateTask (#593)
- [x] AddSubTask (#593)
- [x] RemoveSubTask (#593)
- [x] UpdateSubTask (#593)
- [x] SubTaskToToDo (#593)
- [x] SubTaskToDone (#593)
- [x] TaskToToDo (#593)
- [x] TaskToInProgress (#593)
- [x] TaskToDone (#593)
- [x] AssignTaskToUser (#593)
- [x] UnassignTaskFromUser (#593)
- [x] PublishTask (#593)
[...]

Issue #572, opened 2023-12-03
Ensure consistent ORDER for all requests


FoseFx:
### Describe the chore

task_svc
- [ ] GetWards
- [ ] GetRecentWards
- [ ] GetWardOverviews
- [ ] GetWardDetails
- [ ] GetRoom
- [ ] GetRooms
- [ ] GetRoomsByWard
- [ ] GetRoomOverviewsByWard
- [ ] GetPatientsByWard
- [ ] GetPatientAssignmentByWard
- [ ] GetRecentPatients
- [ ] GetPatientDetails
- [ ] GetPatientList
- [ ] GetAllTaskTemplates
- [ ] GetAllTaskTemplatesByCreator
- [ ] GetAllTaskTemplatesByWard
- [ ] GetTask
- [ ] GetTasksByPatient
- [ ] GetTasksByPatientSortedByStatus
- [ ] GetAssignedTasks
- [ ] BulkCreateBeds
- [ ] GetBeds
- [ ] GetBedsByRoom

user_svc
- [ ] ReadSelf
- [ ] GetOrganization
- [ ] GetOrganizationsByUser
- [ ] GetOrganizationsForUser
- [ ] GetMembersByOrganization
- [ ] GetInvitationsByOrganization
- [ ] GetInvitationsByUser

PR #630, opened 2024-02-07
TaskTemplate Tests

FoseFx:
Progresses #569

Note, that `GetAllTaskTemplatesByCreator` despite its name, does not return all TaskTemplates for a creator. 
That's why some tests make little sense right now. This will be fixed in https://github.com/helpwave/services/pull/625


PaulKalho (approved):
The tests look good. 

Somehow when testing tests no .12 and 29 failed. 
Probably caused by #581 ([12](https://github.com/helpwave/services/pull/581/commits/419ffe72ff7c68d939903563071d2357564af162) and [29](https://github.com/helpwave/services/pull/581/commits/8bb131e3d3eb3ee86a8d1c679b4de4973f98309c)), own issue?

FoseFx:
Tests are flaky, as we nearly never enforce an ORDER for the database, so technically could get back the data in any order. There already is an [issue for that](https://github.com/helpwave/services/issues/572), but it might require some coordination with frontend and medical.

PR #640, opened 2024-02-11
feat(hwes): add projections

### LINKED ISSUE #621: Add projections to `hwes`

MaxSchaefer:
closes #621


libs/common/grpc.go
```diff
@@ -40,7 +40,7 @@ type organizationIDKey struct{}
 //		api.RegisterMyServiceServer(grpcServer, &myServiceServer{})
 //	})
 //	// cleanup after yourself here
-func StartNewGRPCServer(addr string, registerServerHook func(*daprd.Server)) {
+func StartNewGRPCServer(ctx context.Context, addr string, registerServerHook func(*daprd.Server)) {
 	log := zlog.Logger
```
FoseFx, line 44:
```suggestion
	log := zlog.Ctx(ctx)
```

libs/hwes/event.go
```diff
@@ -143,3 +144,10 @@ func (e *Event) SetJsonData(data interface{}) error {
 func (e *Event) GetJsonData(data interface{}) error {
 	return json.Unmarshal(e.Data, data)
 }
+
+func (e *Event) Log(l *zerolog.Event) *zerolog.Event {
```
FoseFx, line 148: confusing name, this method does not log anything

```suggestion
func (e *Event) GetLogEvent(l *zerolog.Event) *zerolog.Event {
```

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+func (p *CustomProjection) Subscribe(ctx context.Context) error {
+	log := zlog.Ctx(ctx)
+
+	// Create subscription on EventStoreDB
+	persistentAllSubscriptionOptions := esdb.PersistentAllSubscriptionOptions{}
+	err := p.es.CreatePersistentSubscriptionAll(ctx, subscriptionGroupName, persistentAllSubscriptionOptions)
+	if err != nil {
+		// Ignore subscription already exists error
+		if persistentSubscriptionError, ok := err.(*esdb.PersistentSubscriptionError); !ok || ok && (persistentSubscriptionError.Code != 6) {
+			zlog.Ctx(ctx).Warn().Err(err).Send()
```
FoseFx, line 53:
```suggestion
			log.Warn().Err(err).Msg("ignoring subscription already exists error")
```

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (p *CustomProjection) Subscribe(ctx context.Context) error {
+	log := zlog.Ctx(ctx)
+
+	// Create subscription on EventStoreDB
+	persistentAllSubscriptionOptions := esdb.PersistentAllSubscriptionOptions{}
+	err := p.es.CreatePersistentSubscriptionAll(ctx, subscriptionGroupName, persistentAllSubscriptionOptions)
+	if err != nil {
+		// Ignore subscription already exists error
+		if persistentSubscriptionError, ok := err.(*esdb.PersistentSubscriptionError); !ok || ok && (persistentSubscriptionError.Code != 6) {
```
FoseFx, line 52: can we use a constant instead of the `6`?

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (p *CustomProjection) Subscribe(ctx context.Context) error {
+	log := zlog.Ctx(ctx)
+
+	// Create subscription on EventStoreDB
+	persistentAllSubscriptionOptions := esdb.PersistentAllSubscriptionOptions{}
+	err := p.es.CreatePersistentSubscriptionAll(ctx, subscriptionGroupName, persistentAllSubscriptionOptions)
+	if err != nil {
+		// Ignore subscription already exists error
```
FoseFx, line 51: why is this ok to do and why are we doing it? (add a comment)

will this cause race-conditions?

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (p *CustomProjection) handleEvent(ctx context.Context, event hwes.Event) (error, esdb.Nack_Action) {
+	eventHandler, found := p.eventHandlers[event.EventType]
+	if !found {
+		return fmt.Errorf("event type '%s' is invalid", event.EventType), esdb.Nack_Unknown
+	}
+	return eventHandler(ctx, event)
+}
+
+func (p *CustomProjection) Subscribe(ctx context.Context) error {
```
FoseFx, line 44: do we need to unsubscribe?

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+	log := zlog.Ctx(ctx)
+
+	// Create subscription on EventStoreDB
+	persistentAllSubscriptionOptions := esdb.PersistentAllSubscriptionOptions{}
+	err := p.es.CreatePersistentSubscriptionAll(ctx, subscriptionGroupName, persistentAllSubscriptionOptions)
+	if err != nil {
+		// Ignore subscription already exists error
+		if persistentSubscriptionError, ok := err.(*esdb.PersistentSubscriptionError); !ok || ok && (persistentSubscriptionError.Code != 6) {
+			zlog.Ctx(ctx).Warn().Err(err).Send()
+			return err
```
FoseFx, line 54: prepend something so we can trace back the error to this line, same for the other error returns
e.g., "projection subscription error:"

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	// Subscribe to server side subscription
```
FoseFx, line 58: not entirely sure what's going on from here on in this function, can you add a short comment, maybe with a link to the relevant docs

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	defer stream.Close()
+
+	log.Info().Msg("Start custom projection")
+	err = p.process(ctx, stream)
+	log.Info().Msg("Stop custom projection")
+	return err
+}
+
+func (p *CustomProjection) process(ctx context.Context, stream *esdb.PersistentSubscription) error {
```
FoseFx, line 71: what does this function process? the name could be more descriptive, and/or add a short comment

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+
+		select {
+		case <-ctx.Done():
+			// Return when context gets finished
+			return nil
+		default:
+		}
+
+		if esdbEvent.SubscriptionDropped != nil {
+			log.Error().Err(esdbEvent.SubscriptionDropped.Error).Msg("Subscription dropped")
```
FoseFx, line 85: not sure how the type of esdbEvent looks like, can you check if there is more information to log (thinking about an event id or alike)?

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+			return esdbEvent.SubscriptionDropped.Error
+		}
+
+		if esdbEvent.EventAppeared == nil || esdbEvent.EventAppeared.Event == nil {
+			log.Debug().Msg("Received empty event, skip")
+			continue
+		}
+
+		if strings.HasPrefix(esdbEvent.EventAppeared.Event.EventType, "$") {
+			// Skip internal events
```
FoseFx, line 95: might be worth logging

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+				Msg("could not create new event from recorded event")
+			continue
+		}
+
+		event.Log(log.Debug()).Msg("process event")
+
+		if err, nackAction := p.handleEvent(ctx, event); err != nil {
+			event.Log(log.Error()).Err(err).Msg("error during processing of event")
+
+			event.Log(log.Warn()).Msg("nack event")
```
FoseFx, line 115: using zerlog.Dict() (or zlog.Dict(), not sure) might be more straight forward

services/tasks-svc/main.go
```diff
@@ -14,12 +17,23 @@ const ServiceName = "tasks-svc"
...
 	eventStore := eventstoredb.SetupEventStoreByEnv()
 	aggregateStore := eventstoredb.NewAggregateStore(eventStore)
 
-	common.StartNewGRPCServer(common.ResolveAddrFromEnv(), func(server *daprd.Server) {
+	echoProjection := echo.NewProjection(eventStore)
+	go func() {
+		err := echoProjection.Base.Subscribe(ctx)
+		if err != nil {
+			log.Err(err).Msg("error during echo subscription")
+			cancel()
```
FoseFx, line 32: cancel must probably also be called after StartNewGRPCServer has finished (see StartNewGRPCServer's comment)

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,135 @@
...
+	zlog "github.com/rs/zerolog/log"
+	"hwes"
+	"strings"
+)
+
+const subscriptionGroupName = "custom"
+
+type eventHandler func(ctx context.Context, evt hwes.Event) (error, esdb.Nack_Action)
+
+type CustomProjection struct {
```
FoseFx, line 16: needs some documentation

services/tasks-svc/internal/task/projections/echo/echo.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"github.com/EventStore/EventStore-Client-Go/esdb"
+	zlog "github.com/rs/zerolog/log"
+	"hwes"
+	"hwes/eventstoredb/projections/custom"
+	taskEventsV1 "tasks-svc/internal/task/events/v1"
+)
+
+type Projection struct {
+	Base *custom.CustomProjection
```
FoseFx, line 13: have you considered struct embedding?

services/tasks-svc/internal/task/projections/echo/echo.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func NewProjection(es *esdb.Client) *Projection {
+	p := &Projection{}
+	p.Base = custom.NewCustomProjection(es)
```
FoseFx, line 18:
```suggestion
	p := &Projection{
		Base: custom.NewCustomProjection(es),
	}
```

services/tasks-svc/internal/task/projections/echo/echo.go
```diff
@@ -0,0 +1,38 @@
...
+	p.Base.
+		RegisterEventListener(taskEventsV1.TaskCreated, p.onTaskCreated)
+}
+
+func (p *Projection) onTaskCreated(ctx context.Context, evt hwes.Event) (error, esdb.Nack_Action) {
+	log := zlog.Ctx(ctx)
+
+	var payload taskEventsV1.TaskCreatedEvent
+	if err := evt.GetJsonData(&payload); err != nil {
+		return err, esdb.Nack_Retry
```
FoseFx, line 33: log this

services/tasks-svc/internal/task/projections/echo/echo.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	log.Info().Str("taskID", payload.ID).Msg("task created, emitted from echo projection")
```
FoseFx, line 36: there is nothing happening in this projection

services/tasks-svc/main.go
```diff
@@ -14,12 +17,23 @@ const ServiceName = "tasks-svc"
...
 	common.Setup(ServiceName, Version, true)
 
 	eventStore := eventstoredb.SetupEventStoreByEnv()
 	aggregateStore := eventstoredb.NewAggregateStore(eventStore)
 
-	common.StartNewGRPCServer(common.ResolveAddrFromEnv(), func(server *daprd.Server) {
+	echoProjection := echo.NewProjection(eventStore)
+	go func() {
+		err := echoProjection.Base.Subscribe(ctx)
+		if err != nil {
```
FoseFx, line 30:
```suggestion
		if err := echoProjection.Base.Subscribe(ctx); err != nil {
```

libs/hwes/eventstoredb/projections/custom/custom.go
FoseFx: also add a comment mentioning this method is thread blocking until the context finishes

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: No race-conditions. If the initial creation fails, we fail on connection as well.

services/tasks-svc/internal/task/projections/echo/echo.go
MaxSchaefer: This empty projection is an example and fully indented to be empty and will get overridden soon

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: Not after the empty-event check. I've added more info to the internal-event check.

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: Very nice! Didn't know `zerolog.Dict()` 👍

services/tasks-svc/internal/task/projections/echo/echo.go
MaxSchaefer: We should eventually introduce a logging middleware for projections to inject more context like the subscribed event?!

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: We are closing the stream at https://github.com/helpwave/services/pull/640/files#diff-0cf98f741b76c7e905a897c32947f1036fd56bc1a01debcba8e343b4435a1e79R76. A explicit unsubscribe is not needed. The state of the client is managed by the server.

services/tasks-svc/internal/task/projections/echo/echo.go
MaxSchaefer: Good call. Thanks!

MaxSchaefer:
@FoseFx I've pinned some `bufbuild/buf-setup-action` versions in https://github.com/helpwave/services/pull/640/commits/7b5f427e66414d23881ab74ad700054529a04a18 and https://github.com/helpwave/services/pull/640/commits/5de16f56ee020d0bda003510d3a8ad48513763e6

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,184 @@
...
+	for {
+		esdbEvent := stream.Recv()
+
+		select {
+		case <-ctx.Done():
+			// Return when context gets finished
+			return nil
+		default:
+		}
+
```
FoseFx, line 130: this is a good point to start a new span

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/EventStore/EventStore-Client-Go/esdb"
+	zlog "github.com/rs/zerolog/log"
+	"hwes"
+	"strings"
+)
+
+const subscriptionGroupName = "custom"
+
+// PersistentSubscriptionFailedCreationErrorCode https://github.com/EventStore/EventStore-Client-Go/blob/v1.0.2/esdb/types.go#L95
+const PersistentSubscriptionFailedCreationErrorCode = 6
```
FoseFx, line 15: when upgrading to v3, try to get rid of this in favor of existing constants

libs/hwes/eventstoredb/projections/custom/custom.go
FoseFx: What will happen when two services are subscribing after another?
Both at connect-time and at event-handle-time

Won't both try to handle the same event at the same time?

services/tasks-svc/main.go
```diff
@@ -14,14 +17,27 @@ const ServiceName = "tasks-svc"
...
 func main() {
+	ctx, cancel := context.WithCancel(context.Background())
+
 	common.Setup(ServiceName, Version, true)
 
 	eventStore := eventstoredb.SetupEventStoreByEnv()
 	aggregateStore := eventstoredb.NewAggregateStore(eventStore)
 
-	common.StartNewGRPCServer(common.ResolveAddrFromEnv(), func(server *daprd.Server) {
+	echoProjection := echo.NewProjection(eventStore)
```
FoseFx, line 27: you can't really do anything else than subscribe to a projection, right?
In that case, it's easy to clean up the code by moving the construction into the go func() block

services/tasks-svc/main.go
```diff
@@ -14,14 +17,27 @@ const ServiceName = "tasks-svc"
...
+	go func() {
+		if err := echoProjection.Subscribe(ctx); err != nil {
+			log.Err(err).Msg("error during echo subscription")
+			cancel()
+		}
+	}()
+
+	common.StartNewGRPCServer(context.Background(), common.ResolveAddrFromEnv(), func(server *daprd.Server) {
```
FoseFx, line 35: We don't share the same context here.

This means the service will continue to run the grpc server when the subscription of the projection fails.
I don't think this is a good idea, the service should either run, and then do so properly, or fail at startup.

Btw we are good the other way around I think.

MaxSchaefer: Yes, you are right. This was my intention with the refactor of `StartNewGRPCServer()` ...

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: Yes. Required a refactor. Done.

libs/hwes/eventstoredb/projections/custom/custom.go
MaxSchaefer: Thats why we are using [Persistent subscriptions](https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html) as well. Every projection gets grouped by its `subscriptionGroupName` (after my refactor from today). Every group has a at-least-once guarantee.

libs/hwes/eventstoredb/projections/custom/custom.go
```diff
@@ -0,0 +1,204 @@
...
+
+// EventStoreDBInternalEventPrefix https://developers.eventstore.com/server/v23.10/streams.html#reserved-names
+const EventStoreDBInternalEventPrefix = "$"
+
+type eventHandler func(ctx context.Context, evt hwes.Event) (error, esdb.Nack_Action)
+
+// CustomProjection can be used to develop own projections
+// A projection is an event sourcing pattern to build up
+// a read model based on the underlying event data.
+// IMPORTANT: The passed subscriptionGroupName identifies you subscription
```
FoseFx, line 24: From what I could gather from my (admittedly brief) look at [the docs](https://developers.eventstore.com/server/v23.10/persistent-subscriptions.html#concepts), group names are not meant to identify subscriptions. In contrary, they are used to batch subscriptions together, and lets the server know which subscribers a message can be load-balanced between.

FoseFx (approved):

PR #655, opened 2024-02-19
Bump deps

FoseFx:


MaxSchaefer (approved):

MaxSchaefer:
Whats up with renovate?

FoseFx:
Renovate works fine, it did pick up these updates. It just bundled them together with others dependencies which we should monitor more carefully, because they are extremely new, for example.

PR #665, opened 2024-02-20
update go-grpc-middleware to v2

FoseFx:


MaxSchaefer (approved):
We are fully compatible with the new major?! 🎉

PR #674, opened 2024-02-29
implements create and getProperty

PaulKalho:
## Which issues does this pull request close?
Progresses #644

Issue #644, opened 2024-02-16
CRUD Properties (as in: Schema)


FoseFx:
API Design: https://github.com/helpwave/services/pull/619/files#diff-44d781779360227a6da04e094379b746ee6d86c958d265028d72ad1e329991e6
and https://github.com/helpwave/services/pull/619/files#diff-294a770289827d8511107f61c3e1cef603a08c64762bb9ae0bfd4bb2ee8b08ec

Also see https://github.com/helpwave/services/issues/611

PR #677, opened 2024-03-01
implements updateProperty

PaulKalho:
progresses #644
stacked on #674


services/property-svc/internal/property/aggregate/aggregate.go
```diff
@@ -101,3 +110,115 @@ func (a *PropertyAggregate) onSetIDUpdated(evt hwes.Event) error {
...
+	return nil
+}
+
+func (a *PropertyAggregate) onFieldTypeDataSelectOptionsUpserted(evt hwes.Event) error {
+	var payload propertyEventsV1.FieldTypeDataSelectOptionsUpsertedEvent
+	if err := evt.GetJsonData(&payload); err != nil {
+		return err
+	}
+
+	if a.Property.FieldTypeData.SelectData != nil {
```
FoseFx, line 185: Here we have one of those race-conditions, which I was worried will happen when we event source too much. Say we want to change a Properties Field Type from Text to Select. Then we emit both "UpdateFieldType" and, "UpsertSelectOptions" events. The order of which is not guaranteed iirc (@MaxSchaefer right?). This means if this event gets processed first, this if is false, and the event silently dropped. If the other event makes it first, we actually commit the new options.

services/property-svc/internal/property/api/grpc.go
```diff
@@ -113,3 +113,54 @@ func (s *PropertyGrpcService) GetProperty(ctx context.Context, req *pb.GetProper
...
+	}
+
+	var none *bool
+	var allowFreetext *bool
+	var removeOptions *[]string
+	var upsertOptions *[]models.SelectOption
+
+	switch ftData := req.FieldTypeData.(type) {
+	case *pb.UpdatePropertyRequest_None:
+		none = &ftData.None
```
FoseFx, line 135: none is not a value, throw everything related to it out (except in the protos ofc)

services/property-svc/internal/property/aggregate/aggregate.go
PaulKalho: If we actually face these issues I could just break down all the Events related to the `Property.FieldTypeData` down to something like `PropertyFieldTypeDataUpdatedEvent`

MaxSchaefer: The order is which the events are applied to the aggregate is guaranteed.
We are talked about the optimistic concurrency. If another aggregate gets loaded and committed to the event store while another request is in flight, the request that commits with the now outdated aggregate version gets rejected.
 
Bildschirmfoto 2024-03-14 um 12 12 07 AM

proto/services/property_svc/v1/property_svc.proto
```diff
@@ -77,19 +77,62 @@ message GetPropertyResponse {
...
+  message SelectData {
+    optional bool allow_freetext = 1;
+    repeated string remove_options = 2; // @gotags: validate:"dive,uuid4"
+    repeated SelectOption upsert_options = 3;
+
+    message SelectOption {
+      string id = 1;
+      optional string name = 2;
+      optional string description = 3;
+      // TODO: TBD: optional bool is_custom = 4; why is that only in updateProperty and not in createProperty?
```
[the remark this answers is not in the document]
PaulKalho: We should also return `is_custom` in `GetPropertyResponse` ?

services/property-svc/internal/property/models/property.go
```diff
@@ -15,7 +15,7 @@ type Property struct {
 	IsArchived  bool
 
 	SetID         uuid.NullUUID
-	FieldTypeData FieldTypeData
+	FieldTypeData *FieldTypeData
```
FoseFx: Properties have FieldTypeData. Always. The FieldTypeData itself might be full of only nil values though. Here we have a similar issue as we had with SelectOption. I think we should have a seperate UpdateProperty model, for which FieldTypeData may be nil, when it should not be changed

services/property-svc/internal/property/aggregate/aggregate.go
```diff
@@ -182,15 +184,56 @@ func (a *PropertyAggregate) onFieldTypeDataSelectOptionsUpserted(evt hwes.Event)
...
 	}
 
-	if a.Property.FieldTypeData.SelectData != nil {
-		// TODO: Check and update already existing SelectOptions don't just append them
-		a.Property.FieldTypeData.SelectData.SelectOptions = append(a.Property.FieldTypeData.SelectData.SelectOptions, payload.UpsertedSelectOptions...)
-	} else {
-		a.Property.FieldTypeData.SelectData = &models.SelectData{
-			SelectOptions: payload.UpsertedSelectOptions,
+	var initialAllowFreetext bool
+	existingSelectOptions := make(map[uuid.UUID]models.SelectOption)
```
FoseFx, line 196: We update this value later on, at which point the name `existingSelectOptions` is not appropriate anymore

services/property-svc/internal/property/aggregate/aggregate.go
```diff
@@ -182,15 +184,56 @@ func (a *PropertyAggregate) onFieldTypeDataSelectOptionsUpserted(evt hwes.Event)
...
+				updatedOpt.IsCustom = *option.IsCustom
+			}
+			existingSelectOptions[option.ID] = updatedOpt
 		}
 	}
 
+	var updatedSelectOptions []models.SelectOption
+	for _, opt := range existingSelectOptions {
+		updatedSelectOptions = append(updatedSelectOptions, opt)
+	}
```
FoseFx, line 233: We know the size of the new array. AFAIK it's way more performant to avoid append here.

services/property-svc/internal/property/aggregate/aggregate.go
```diff
@@ -182,15 +184,56 @@ func (a *PropertyAggregate) onFieldTypeDataSelectOptionsUpserted(evt hwes.Event)
...
+
+	a.Property.FieldTypeData = &models.FieldTypeData{
```
FoseFx: can we, instead of creating a new FieldTypeData struct, update the existing one?
Say we add another value to the struct later on, then it would be removed here.

services/property-svc/internal/property/aggregate/aggregate.go
PaulKalho: You're right, thanks for that!

services/property-svc/internal/property/models/property.go
PaulKalho: I don't see the need for an own UpdatePropertyModel we are not facing the similar issue as with SelectOption if I understand correctly.

PaulKalho: I get why it is semantically wrong, that's why i fixed that in my latest commit.
> we are not facing the similar issues

PaulKalho: I get why it is semantically wrong, that's why i fixed that in my latest commit.
> we are not facing the similar issues 

With that I just wanted to express that we don't need an extra `UpdateProperty` model atm where `FieldTypeData` is optional, all update operations on SelectData (upsert, remove, allowFreetext) are taken care of.

FoseFx: but how do we model no change to FieldTypeData?

PaulKalho: `FieldTypeData` is of type oneof an thus dont has to be set in the request.
If`FieldTypeData` is i.e `SelectData` in the request the operations (upsert, remove, allowFreeText) will trigger the according Events and changes will be applied to the aggregate. 

https://github.com/helpwave/services/blob/daefb51f3e63a06d9973ccf381d33a81cc8a510c/services/property-svc/internal/property/api/grpc.go#L113

If FieldTypeData is not set, no events will be triggerted

FoseFx (approved):

PR #683, opened 2024-03-03
feat: allow the UserID to be injected into events

### LINKED ISSUE #682: Inject the users identity into our events
### Describe the chore

We should store the users identity via their ID in the our EventStore as well to track the actual caller.

### LINKED ISSUE #695: `resolveAggregateIDAndTypeFromStreamID` returns `(type, id)` instead of `(id, type)`
### Describe the chore

Might cause bugs

MaxSchaefer:
closes #682
closes #695


dapr/config.yaml
```diff
@@ -7,4 +7,4 @@ spec:
     component: "kubernetes"
     configuration:
       # https://fly.io/docs/reference/private-networking/#fly-internal-addresses
-      template: "helpwave-staging-{{.ID}}.internal:{{.Port}}" # "helpwave-staging" will be replaced by the actual organization in fly-deploy.yaml
+      template: "helpwave-production-{{.ID}}.internal:{{.Port}}" # "helpwave-staging" will be replaced by the actual organization in fly-deploy.yaml
```
FoseFx, line 10: ?

libs/hwes/event.go
```diff
@@ -18,6 +21,11 @@ type Event struct {
 	Data          []byte
 	Timestamp     time.Time
 	Version       uint64
+	UserID        *uuid.UUID
+}
+
+type metadata struct {
+	UserID string `json:"user_id"`
```
FoseFx, line 28: Who's user id? The name does not reflect, that we mean the "committer" or something

libs/hwes/event.go
```diff
@@ -40,6 +49,28 @@ func NewEventWithData(aggregate Aggregate, eventType string, data interface{}) (
 	return event, nil
 }
 
+// EventWithUserID injects the UserID from the passed context via common.GetUserID().
+// If no UserID was injected, prior to this function call, an error will be returned.
+// Make sure to inject the UserID via a Middleware in the API layer.
+func EventWithUserID(ctx context.Context, event Event) (Event, error) {
```
FoseFx, line 55: Make this the "normal" Event constructor, and rename the other to "EventWithoutUserID", as most events (if not all?) are initiated by a user.

FoseFx: maybe use the options pattern instead of creating multiple constructors?

libs/hwes/event.go
MaxSchaefer: https://github.com/helpwave/services/issues/682
> We should store the users identity via their ID in the our EventStore as well to track the actual caller.

Thats then more of a general discussion. The goal is to track the identity that is directly responsible for this event.
- `UserID` assumes that an event can be triggered by a user. Thats true (for now).
- `CommitterID` would move the whole user/identity concept out of `hwes`. 
- `IdentityID` would bind more loosely these two concepts, an event and a user/identity. But then, its not clear, right now, if the `IdentityID` represents a user or for example some sort of service account.

I recommend: For the metadata, `CommiterID` should be good but gets then translated to `UserID` in `hwes.Event`.

libs/hwes/event.go
```diff
@@ -49,10 +59,25 @@ func NewEventWithData(aggregate Aggregate, eventType string, data interface{}) (
...
+	}
+
+	return event, err
+}
+
+// InjectUserIDInEventFromContext injects the UserID from the passed context via common.GetUserID().
 // If no UserID was injected, prior to this function call, an error will be returned.
 // Make sure to inject the UserID via a Middleware in the API layer.
-func EventWithUserID(ctx context.Context, event Event) (Event, error) {
+func InjectUserIDInEventFromContext(ctx context.Context, event Event) (Event, error) {
```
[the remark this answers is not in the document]
FoseFx: Still not sure why we don't just change the `NewEvent` function.

Either to `func NewEvent(aggregate Aggregate, eventType string, ctx context.Context) (Event, error)` or to something extensible, so we can get rid of `NewEventWithUser` and `NewEventWithData` as well using the options pattern:

```go

type EventOption func(*Event) error

// WithData sets the data payload for the event, must be json-marshable
func WithData(data interface{}) EventOption {
    return func (event *Event) {
        return event.SetJsonData(&data)
    }
}

// WithContext adds the context's user id as a committer for the event
func WithContext(ctx context.Context) EventOption {
    return func (event *Event) {
       _, err := event.SetCommitterFromCtx(ctx)
	   return err
    }
}

//     event, err := NewEvent(aggregate, "event-type", WithContext(ctx), WithData(&somePayload))
func NewEvent(aggregate Aggregate, eventType string, opts EventOption...) (Event, error) {
  ...
  for _, option := range opts {
     err := opts(&event)
     if err != nil { return Event{}, err }
  }
  ...
}

```

Anything else is not scalable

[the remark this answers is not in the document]
MaxSchaefer: Didn't thought about the options pattern. Looks perfect for that, thanks!

MaxSchaefer:
fyi: I commented out all `tracking` code due to https://github.com/helpwave/services/issues/694

libs/hwes/event.go
```diff
@@ -18,32 +21,51 @@ type Event struct {
 	Data          []byte
 	Timestamp     time.Time
 	Version       uint64
+	UserID        *uuid.UUID
```
FoseFx, line 24: We now have both `Event.UserID` and `metadata.CommitterID`.

libs/hwes/event.go
```diff
@@ -18,32 +21,51 @@ type Event struct {
...
-func NewEvent(aggregate Aggregate, eventType string) Event {
-	return Event{
+type metadata struct {
+	// CommitterId represents some sort of optional identity that is directly responsible for this event (e.g., a user id). Must be a uuid
+	CommitterID string `json:"committer_id"`
+}
+
+type EventOption func(*Event) error
+
+func WithContext(ctx context.Context) EventOption {
```
FoseFx, line 34:
```suggestion
// WithContext applies SetCommitterFromCtx after construction
func WithContext(ctx context.Context) EventOption {
```

libs/hwes/event.go
```diff
@@ -18,32 +21,51 @@ type Event struct {
...
+	return func(event *Event) error {
+		return event.SetCommitterFromCtx(ctx)
+	}
+}
+
+func WithData(data interface{}) EventOption {
```
FoseFx, line 40:
```suggestion
// WithData applies SetJsonData after construction
func WithData(data interface{}) EventOption {
```

libs/hwes/event.go
```diff
@@ -75,26 +97,38 @@ func resolveAggregateIDAndTypeFromStreamID(streamID string) (aggregateType Aggre
...
 		AggregateID:   aggregateID,
 		AggregateType: aggregateType,
-		Data:          event.Data,
-		Timestamp:     event.CreatedDate,
-		Version:       event.EventNumber,
-	}, nil
+		Data:          esdbEvent.Data,
+		Timestamp:     esdbEvent.CreatedDate,
+		Version:       esdbEvent.EventNumber,
+	}
```
FoseFx, line 124: for good measure, have an explicit `UserID: nil` here

libs/hwes/event.go
FoseFx: at least rename UserID

FoseFx (approved):

PR #708, opened 2024-03-28
setup projections for property svc

### LINKED ISSUE #707: Setup projections for `property-svc`
### Describe the feature

Similar to https://github.com/helpwave/services/pull/702, just for the `property-svc`

As a first query that consumes these projections, we need a query like "List all properties for edit by subject type".

### Additional context

_No response_

PaulKalho:
closes #707 #644


PaulKalho:
@MaxSchaefer @FoseFx no idea why the staging version is newer than this one... Also I don't get why the linter is complaining. `go mod tidy` does not change a thing.

MaxSchaefer:
- Merged main
- `go mod tidy` for `property-svc` afterwards
- `./models.sh property-svc` cause of sqlc update
- `helpwave_staging_property_svc` database was pre-filled with data from initial test-deployment, I've dropped the whole database
- Seams like that our migration-check CI does not handle initial migrations correctly (because down migrations does not exist)

FoseFx:
> Seams like that our migration-check CI does not handle initial migrations correctly, ./migration.sh property-svc goto 1 fails

The `goto 0` is the problem. I opened #724 for this.

Here is a run with the fix, on a squashed version of this PR: https://github.com/helpwave/services/actions/runs/8587412094/job/23531021892

In fact, there is an issue with the down migration.

services/property-svc/migrations/000001_init.up.sql
```diff
@@ -0,0 +1,45 @@
...
+	name text NOT NULL,
+	description text NOT NULL DEFAULT '',
+	is_custom boolean NOT NULL DEFAULT false,
+	select_data_id uuid NOT NULL,
+	FOREIGN KEY (select_data_id)
+		REFERENCES select_datas(id)
+		ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS field_type_datas (
```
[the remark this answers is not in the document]
PaulKalho: Yes I also thought about that. 
I decided to put that into an extra table because I wanted to model our db representation as close as possible to our event sourced structure, and I think it's not too much of an overhead.
Do you think I should get rid of it? @FoseFx

services/property-svc/migrations/000001_init.up.sql
```diff
@@ -0,0 +1,45 @@
...
+	id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+	select_data_id uuid DEFAULT NULL,
+	-- check: oneof
+	check(
+		(
+			(select_data_id is not null)::integer
+		) <= 1
+	),
```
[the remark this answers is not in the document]
PaulKalho: I left it out for now, didn't make so much sense in the init migration. If we add more field_type_datas we should add this constraint again.

services/property-svc/migrations/000001_init.up.sql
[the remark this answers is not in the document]
FoseFx: For field types that do not need additional data we would need to maintain a field data entry anyway. Also, we constantly need to `JOIN` the two tables, which takes a lot of computation to do ig.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+		return fmt.Errorf("field_type %s invalid", payload.FieldType), esdb.NackActionRetry
+	}
+	fieldType := (pb.FieldType)(value)
+
+	err = propertyRepo.CreateProperty(ctx, property_repo.CreatePropertyParams{
+		ID:              propertyID,
+		SubjectType:     int32(subjectType),
+		FieldType:       int32(fieldType),
+		Name:            payload.Name,
+		Description:     "",
```
FoseFx, line 101: do we not get the description here? we should, right? (change proto if needed)

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+		if !errors.Is(err, pgx.ErrTxClosed) {
+			log.Error().Err(err).Msg("rollback failed.")
+		}
+	}()
+
+	propertyRepo := property_repo.New(db).WithTx(tx)
+
+	var payload propertyEventsV1.PropertyCreatedEvent
+	if err := evt.GetJsonData(&payload); err != nil {
+		log.Error().Err(err).Msg("unmarshal failed")
```
FoseFx, line 68: payload might be interesting

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+	log := zlog.Ctx(ctx)
+
+	var payload propertyEventsV1.PropertySetIDUpdatedEvent
+	if err := evt.GetJsonData(&payload); err != nil {
+		log.Error().Err(err).Msg("unmarshal failed")
+		return err, esdb.NackActionRetry
+	}
+
+	setID := uuid.NullUUID{UUID: uuid.Nil, Valid: false}
+	if len(payload.SetID) > 0 {
```
FoseFx, line 147:
```suggestion
	if payload.SetID != "" {
```

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+		parsedID, err := hwutil.ParseNullUUID(&payload.SetID)
+		if err != nil {
+			return err, esdb.NackActionRetry
+		}
+		setID = parsedID
```
FoseFx, line 152:
```suggestion
	if parsedID, err := hwutil.ParseNullUUID(&payload.SetID); err != nil {
		return err, esdb.NackActionRetry
	} else {
		setID = parsedID
	}
```

FoseFx: how would this handle the case where a property is removed from a set?

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+
+	value, found := pb.SubjectType_value[payload.SubjectType]
+	if !found {
+		return fmt.Errorf("invalid fieldType: %s", payload.SubjectType), esdb.NackActionRetry
+	}
+	subjectType := (pb.SubjectType)(value)
+
+	err := p.propertyRepo.UpdateProperty(ctx, property_repo.UpdatePropertyParams{
+		SubjectType: hwutil.PtrTo(int32(subjectType)),
+		ID:          evt.AggregateID,
```
FoseFx, line 184:
```suggestion
		ID:          evt.AggregateID,
		SubjectType: hwutil.PtrTo(int32(subjectType)),
```

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+	if !found {
+		return fmt.Errorf("subject_type %s invalid", payload.SubjectType), esdb.NackActionRetry
+	}
+	subjectType := (pb.SubjectType)(value)
+
+	value, found = pb.FieldType_value[payload.FieldType]
+	if !found {
+		return fmt.Errorf("field_type %s invalid", payload.FieldType), esdb.NackActionRetry
+	}
+	fieldType := (pb.FieldType)(value)
```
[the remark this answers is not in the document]
PaulKalho: No it cannot. We are checking if its "valid" in the lines above.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -0,0 +1,614 @@
...
+	propertyID, err := uuid.Parse(payload.ID)
+	if err != nil {
+		return err, esdb.NackActionRetry
+	}
+
+	value, found := pb.SubjectType_value[payload.SubjectType]
+	if !found {
+		return fmt.Errorf("subject_type %s invalid", payload.SubjectType), esdb.NackActionRetry
+	}
+	subjectType := (pb.SubjectType)(value)
```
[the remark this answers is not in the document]
PaulKalho: No it cannot. We are checking if its "valid" in the lines above.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
PaulKalho: If a property is beeing removed then `payload.SetID == ""`. If this is the case we will update to uuid.NullUUID in the db.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
PaulKalho: If we create a property with a description, which we can, we are dispatching the `PropertyCreated` event as well as the `PropertyDescriptionUpdatedEvent` which then updates the description of the property.
This is exactly how we do it in the task projection as well. I will remove the description as well as the isArchived from the sql statement.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
PaulKalho: I would say the error is enough here... Ofc, I can add it if you'd like. Your decison.

FoseFx (approved):

PR #735, opened 2024-04-16
custom projection: add stream-type filter

### LINKED ISSUE #734: Define event-type in custom projection
### Describe the chore

I think we should have the possibility to define the event-type which the custom projection consumes.

Atm we only have projections which are about one aggregate (e.g `property`, `task`), but the task-projection also consumes property events and always logs `event type ... is invalid`.

PaulKalho:
closes #734 

This also filters out all internal/system events theoretically.
If we wanted to, we could use `esdb.ExcludeSystemEventsFilter(),` ([see](https://developers.eventstore.com/clients/grpc/subscriptions.html#filtering-out-system-events)) as our default filter and throw out: https://github.com/helpwave/services/blob/7f085c302e9e14f857b2a4d8f88f50e3e9b30087/libs/hwes/eventstoredb/projections/custom/custom.go#L159-L165
Wdyt? @MaxSchaefer


FoseFx (approved):

MaxSchaefer:
> ## Which issues does this pull request close?
> closes #734
> 
> This also filters out all internal/system events theoretically. If we wanted to, we could use `esdb.ExcludeSystemEventsFilter(),` ([see](https://developers.eventstore.com/clients/grpc/subscriptions.html#filtering-out-system-events)) as our default filter and throw out:
> 
> https://github.com/helpwave/services/blob/7f085c302e9e14f857b2a4d8f88f50e3e9b30087/libs/hwes/eventstoredb/projections/custom/custom.go#L159-L165
> 
> 
> Wdyt? @MaxSchaefer

Yes, please throw out our custom internal event logic while always applying `esdb.ExcludeSystemEventsFilter()`.

MaxSchaefer (approved):

PR #743, opened 2024-05-02
refactor esdb nack actions

### LINKED ISSUE #727: RFC: Unmarshall errors should not be retried
### Describe the chore

We have numerous snippets like this:
```
if err := evt.GetJsonData(&payload); err != nil {
		return err, esdb.NackActionRetry
	}
```

But if an event can't be processed because it is invalid (syntactically, or schema-wise), it will not be processed the next time around. Only after a significant time has passed, and a developer pushed a change, the event might be handled.
So if a user changes the description of a task and something goes wrong in the Command handler, yielding a bad event, not only is there no error emitted, but the description might change weeks later, possibly overwriting changes made since.

PaulKalho:
closes #727


PaulKalho:
`NackActionRetry` events will be retried 10 times. We could change this to any value in  `CreatePersistentSubscriptionToAll()`

PaulKalho:
@MaxSchaefer I observed that `NackActionUnknown` will be treated like `NackActionRetry` atm. I am not quite sure where/if we can change that behavior.

services/property-svc/internal/property/projections/property_postgres_projection/property_postgres_projection.go
```diff
@@ -417,7 +417,7 @@ func (p *Projection) onFieldTypeDataSelectOptionsUpserted(ctx context.Context, e
 		} else {
 			if selOpt.Name == nil {
 				log.Error().Msg("selectOption name has to be set on create")
-				return nil, esdb.NackActionRetry
+				return nil, hwutil.PtrTo(esdb.NackActionRetry)
```
MaxSchaefer, line 420:
```suggestion
				// existsSelectOptions gets modified by a database result. Thats why we are retrying here.
				return nil, hwutil.PtrTo(esdb.NackActionRetry)
```

MaxSchaefer (approved):
`PtrTo` `PtrTo` `PtrTo` `PtrTo` `PtrTo` `PtrTo` we can probably introduce some sort of own enum later on that maps to esdb 👍

Issue #703, opened 2024-03-20
add missing endpoints in tasks-svc


PaulKalho:
### Describe the chore

Without projections, some endpoints couldn't be realized. These are the missing ones:

task
- [ ] GetTasksByPatient
- [ ] GetTasksByPatientSortedByStatus
- [ ] GetAssignedTasks

patient
- [ ] GetPatientByBed
- [ ] GetPatientsByWard
- [ ] GetPatientAssignmentByWard
- [ ] GetPatientDetails
- [ ] GetPatientList

PR #755, opened 2024-05-17
add patient handlers registry

PaulKalho:
Need that for #703


services/tasks-svc/internal/patient/api/grpc_test.go
```diff
@@ -21,7 +22,9 @@ import (
 func server(ctx context.Context) (pb.PatientServiceClient, func()) {
 	// Build gRPC service
 	aggregateStore := hwes_test.NewAggregateStore()
-	patientGrpcService := api.NewPatientGrpcService(aggregateStore)
+	patientHandlers := handlers.NewPatientHandlers(aggregateStore)
```
FoseFx, line 25: if patientHandlers is not used anywhere else, why not move it into the NewPatientGrpcService constructor?

FoseFx (approved):

PR #794, opened 2024-08-21
refactor pb.date

### LINKED ISSUE #793: Use `google.protobuf.Timestamp` instead of `Date` for `date_value` in `AttachPropertyValueRequest`
### Describe the chore

_See title_
Our own `Date` values comes without any further documentation. We can either document the behaviour of `Date` or we just use `google.protobuf.Timestamp` while ignoring the time part.
https://github.com/helpwave/services/blob/main/proto/services/property_svc/v1/property_value_svc.proto#L21

FoseFx:
closes #793


services/property-svc/internal/property-value/commands/v1/attach_property_value.go
```diff
@@ -16,7 +16,6 @@ func NewAttachPropertyValueCommandHandler(as hwes.AggregateStore) AttachProperty
 		propertyValueRepo := property_value_repo.New(hwdb.GetDB())
 		var a *aggregate.PropertyValueAggregate
 
-		// TODO: Do not use the data from the projection here, could be outdated
```
MaxSchaefer, line 19: Still true

PR #795, opened 2024-08-23
for consistency, use ZED_ envs everywhere

FoseFx:


FoseFx:
Fly secrets are already set

MaxSchaefer:
Where is the inconsistency? I prefer `SPICEDB_` due to clarity. `ZED` is a keyword inside the SpiceDB ecosystem.

FoseFx:
zed cli tool, (and spice.go) use ZED_, so we'd need to set two (actually four to six) environment variables. Maintenance burden.

MaxSchaefer (approved):

PR #803, opened 2024-09-05
chore: add `FEATURE_FLAG_HWAUTHZ_SPICEDB_CHECKS_ALWAYS_TRUE`

MaxSchaefer:
This pull requests adds the feature flag `FEATURE_FLAG_HWAUTHZ_SPICEDB_CHECKS_ALWAYS_TRUE` for `hwauthz`. Currently we do not have a prod/staging ready SpiceDB config.

The feature flag is already set for `tasks-svc`

### DIFF
--- a/libs/hwauthz/README.md
+++ b/libs/hwauthz/README.md
@@ -4,3 +4,9 @@
 
 Provides an SDK for Fine-Grained Authorization using a [Zanzibar](https://zanzibar.academy/)-like provider ([SpiceDB](https://authzed.com/docs/spicedb/getting-started/discovering-spicedb)).
 
+## Feature flags
+
+- `FEATURE_FLAG_HWAUTHZ_SPICEDB_CHECKS_ALWAYS_TRUE`
+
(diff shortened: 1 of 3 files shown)


PaulKalho (approved):

FoseFx:
No? Just don't add checks if we don't have a proper permission to check against. (We can comment out the existing dummy ones if they cause problems). Also such a flag is NOT a feature flag! Instead it must be prefixed with INSECURE_ or alike

FoseFx (changes requested):
^^

MaxSchaefer:
@FoseFx feel free to take this PR over. We need a solution to allow for a short period of time, until proper spicedb schemas are in place, all permission checks. So that our clients can develop further. The intro of SpiceDB was heavily breaking.

MaxSchaefer:
https://github.com/helpwave/services/pull/817 already removes checks. For proper authZ checks, we need to wait for https://github.com/helpwave/services/pull/832

PR #812, opened 2024-09-11
First Property Permissions

FoseFx:
* moves spice.go into cmd/spice
* moves parts of spice.go to hwauthz
* add spicedb to stories tests
* fix testcontainers eventstore issue, which prevented usage of modules
* rename teams to organizations (technically breaking, but never used)
* permissive permissions for CRU(~~D~~) of properties
* integration of those into property-svc
* add index as a first parameter to hwutil.Filter


services/property-svc/internal/property/queries/v1/get_properties_by_subject_type.go
```diff
@@ -84,6 +91,19 @@ func NewGetPropertiesQueryHandler() GetPropertiesQueryHandler {
...
+
+		canGet, err := authz.BulkCheck(ctx, checks)
+		if err != nil {
+			return nil, err
+		}
+
+		properties = hwutil.Filter(properties, func(i int, _ *models.PropertyWithConsistency) bool {
+			return canGet[i]
+		})
+
```
MaxSchaefer, line 106: We should prefer [looking up all resources beforehand](https://buf.build/authzed/api/docs/main:authzed.api.v1#authzed.api.v1.PermissionsService.LookupResources) instead of filtering them afterward. This is also not an consistency issue because when SpiceDB is out of sync, the permission checks will fail.

services/property-svc/internal/property/queries/v1/get_property_by_id.go
```diff
@@ -66,6 +73,12 @@ func NewGetPropertyByIDQueryHandler() GetPropertyByIDQueryHandler {
 			}
 		}
 
+		// Verify user is allowed to see this property
+		check := hwauthz.NewPermissionCheck(user, perm.PropertyCanUserGet, perm.Property(propertyID))
+		if err = authz.Must(ctx, check); err != nil {
+			return nil, 0, err
+		}
+
```
MaxSchaefer, line 81: Move this to the start of the command. We only spend computing if the checks succeeds.

MaxSchaefer (changes requested):
Can you please remove spicedb/team*.{yaml,zed} as well?

services/property-svc/internal/property/queries/v1/get_properties_by_subject_type.go
FoseFx: You are likely to have access to a large amount of properties. The approach you suggest does not make sense in that case.

And equally: We expect to actually filter properties out very rarely (currently, not at all, you can see all properties in your organization).

As the requirements and thus permissions change, it might make more sense to do your approach.

Here is a guide: https://authzed.com/docs/spicedb/modeling/protecting-a-list-endpoint

libs/hwauthz/spicedb/migrate/migrate.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// now collect the first element
+	response, err := stream.Recv()
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			return 0
+		}
+		if statusErr, ok := status.FromError(err); ok {
+			if statusErr.Code() == codes.FailedPrecondition &&
+				statusErr.Message() == "object definition `spice_schema_migrations/migration` not found" {
```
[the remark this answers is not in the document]
FoseFx: yes, from what I can see there is no stable way to get the missing object definition's name.

as for the panic, that's totally fine, as this code is only used in the spice cli tool and test setups.

FoseFx:
It'd be great if we can get to this soon, then I can start with the other permissions without having to rebase many many PRs.

MaxSchaefer (approved):
Awesome 🚀🔓

FoseFx:
> Please check the web frontend after this got merged

Existing things will break, we dont backfill relationships

PR #834, opened 2024-09-17
Optimistic Concurrency Pt.1

### LINKED ISSUE #808: Consistency Tokens / Optimistic Concurrency
### Describe the feature

Consistency Tokens are used to communicate the assumed state of a resource (i.e., aggregate). The last event processed on an aggregate can serve as such a Consistency Token.

Reads should carry consistency-tokens (i.e., event ids) for every aggregate involved.
When a client attempts to write to an aggregate, they should be able to provide this token.

If the state has not changed since, the update is processed. Else, the client is provided with *a more current* state of the aggregate and its event id. As the relation between the services and EventStore (which is the source of truth) is only eventually-consistent, we are not able to guarantee this state is, in fact, *the current* state.

The frontend is can then resolve the conflict, wither automatically, or with help of the user.

Multiple clients may attempt to write to the same aggregate at the same time.
Our endpoints generally only require the changed fields to be sent to the backend, which contains the impact (e.g., if Alice updates the name of a patient, and Bob the same patient's bed, no change will be overwritten).
However, concurrent writes to the same field can of course still occur. Writes themselves are atomic, which means we will never end up in an *invalid* state, yet a slightly delayed write might overwrite a change that was just made. The damage is especially great with large free-text fields, for example. When two people are trying to add a new paragraph, only the last write "wins".

This problem is of course solved. The following alternative solutions have been considered:
* [Conflict-free replicated data types](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type)
* Notify every potential writer about changes (and stream them in, if possible)

Currently, we do not assume any aggregate to suffer from high congestion, thus the solutions above don't seem to reasonable in terms of dev cost and maintenance.

### Additional context

_No response_

FoseFx:
First Steps for #808

- [x] Return Consistency Token
  - [x] CreatePropertyResponse
  - [x] GetPropertyResponse
  - [x] GetPropertiesBySubjectTypeResponse
  - [x] GetAttachedPropertyValuesResponse
  - [x] GetBedResponse
  - [x] GetBedByPatientResponse
  - [x] GetBedsResponse
  - [x] GetBedsByRoomResponse
  - [x] GetPatientResponse
  - [x] GetPatientByBedResponse
  - [x] GetPatientsByWardResponse
  - [x] GetPatientAssignmentByWardResponse
  - [x] GetPatientDetailsResponse
  - [x] GetPatientListResponse
  - [x] GetRecentPatientsResponse
  - [x] GetRoomResponse
  - [x] GetRoomsResponse
  - [x] GetRoomOverviewsByWardResponse
  - [x] GetTaskResponse
  - [x] GetTasksByPatientResponse
  - [x] GetTasksByPatientSortedByStatusResponse
  - [x] GetAssignedTasksResponse
  - [x] GetAllTaskTemplatesResponse
  - [x] GetWardResponse
  - [x] GetWardsResponse
  - [x] GetRecentWardsResponse
  - [x] GetWardDetailsResponse
  - [x] GetWardOverviewsResponse
  - [x] CreateBedRequest
  - [x] BulkCreateBedsResponse (not implemented?)
  - [x] CreatePatientRequest
  - [x] CreateRoomResponse 
  - [x] CreateTaskRequest
  - [x] CreateSubtaskResponse
  - [x] CreateTaskTemplateResponse 
  - [x] CreateTaskTemplateSubTaskResponse 
  - [x] UpdatePropertyResponse
  - [x] AttachPropertyValueResponse
  - [x] UpdateBedResponse
  - [x] UpdatePatientResponse
  - [x] AssignBedResponse
  - [x] UnassignBedResponse 
  - [x] DischargePatientResponse
  - [x] ReadmitPatientResponse
  - [x] UpdateRoomResponse
  - [x] UpdateTaskRequest
  - [x] AssignTaskResponse
  - [x] UnassignTaskResponse
  - [x] UpdateSubtaskResponse
  - [x] RemoveTaskDueDateResponse
  - [x] UpdateTaskTemplateResponse
  - [x] UpdateTaskTemplateSubTaskResponse
- [x] Make sure all projections update the consistency
  - [x] ward_repo.sql
  - [x] bed_repo.sql
  - [x] task_repo.sql
  - [x] room_repo.sql
  - [x] task_template_repo.sql
  - [x] patient_repo.sql
  - [x] property_repo.sql
  - [x] property_value


services/tasks-svc/stories/PatientCRUD_test.go
```diff
@@ -0,0 +1,684 @@
...
+
+	for _, room := range res.Rooms {
+		assert.Contains(t, roomIds, room.Id)
+		assert.Equal(t, roomConsistencies[room.Id], room.Consistency)
+
+		for _, bed := range room.Beds {
+			assert.Contains(t, bedsForRoom[room.Id], bed.Id)
+			assert.Equal(t, bedConsistencies[bed.Id], bed.Consistency)
+
+			if bed.Patient == nil {
```
MaxSchaefer, line 347: `bed.Patient == nil` should never be the case because for every bed a patient gets created, right?

FoseFx: yes, but i'ts not guaranteed by the type system and should therefore be tested

services/property-svc/stories/GetPropertieBySubjectType_test.go
```diff
@@ -0,0 +1,163 @@
...
+		assert.NoError(t, err)
+
+		re := regexp.MustCompile(`[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}`)
+		res := re.ReplaceAllString(string(s), "<id>")
+
+		v := make(map[string]interface{})
+		err = json.Unmarshal([]byte(res), &v)
+		assert.NoError(t, err)
+
+		propertyMap[property.Id] = v
```
MaxSchaefer, line 123: Why removing the IDs?

MaxSchaefer (changes requested):
Not in scope of this pr but whats your opinion about an own type definition for the consistency tokens instead of `uint64`?

MaxSchaefer:
261/261

MaxSchaefer (changes requested):
All touched Proto definitions needs to be backwards compatible. Many Proto changes of this PR are introducing a non-optional param "consistency". This param needs to be optional. 

Examples
-  https://github.com/helpwave/services/pull/834/files#diff-8dd286a5d12ffa4ff06473e9e361f3c375b853e077de8e77277565f94eee9471R34
- https://github.com/helpwave/services/pull/834/files#diff-8dd286a5d12ffa4ff06473e9e361f3c375b853e077de8e77277565f94eee9471R76

services/property-svc/stories/GetPropertieBySubjectType_test.go
FoseFx: because we dont have the subtask ids to compare to

FoseFx:
> All touched Proto definitions needs to be backwards compatible. Many Proto changes of this PR are introducing a non-optional param "consistency". This param needs to be optional.
> 
> Examples
> 
> * https://github.com/helpwave/services/pull/834/files#diff-8dd286a5d12ffa4ff06473e9e361f3c375b853e077de8e77277565f94eee9471R34
> * https://github.com/helpwave/services/pull/834/files#diff-8dd286a5d12ffa4ff06473e9e361f3c375b853e077de8e77277565f94eee9471R76

I just went through all message changes again, requests only got new `optional` fields. Responses got new mandatory strings, because the backend must provide them. (Clients are free to ignore it though)

FoseFx:
> Not in scope of this pr but whats your opinion about an own type definition for the consistency tokens instead of uint64?

Very much in scope, and a very good idea!

MaxSchaefer (approved):
Stay consistent 🤟

PR #841, opened 2024-09-26
Optimistic Concurrency Pt.2

### LINKED ISSUE #808: Consistency Tokens / Optimistic Concurrency

FoseFx:
closes #808

Stacked on #834


FoseFx:
Now blocked by #880

MaxSchaefer (changes requested):
Thats a huge
- Enhancement of conflict detection
- Large rewrite of some types in the property-svc
- Bare bone conflict resolution via retries
- Conflict diffs (most of this pr)

The conflict diff is done manually in our API layer. All of those areas are marked with a TODO comment to find a more generic approach later on. Are there currently any ideas how we can abstract this?

PR #848, opened 2024-10-04
feat: unassign bed on discharge

### LINKED ISSUE #843: Patient Discharge does not unassign from bed
### Describe the bug

Patient Discharge does not unassign from bed

1. Discharge a Patient assigned to a Bed
=> Patient is discharged but still assigned to a Bed

### Additional context

Occures on the tasks-service

PaulKalho:
closes #843

### DIFF
--- a/services/tasks-svc/internal/patient/aggregate/actions.go
+++ b/services/tasks-svc/internal/patient/aggregate/actions.go
@@ -26,12 +26,7 @@ func (a *PatientAggregate) AssignBed(ctx context.Context, newBedID uuid.UUID) er
 	return a.Apply(event)
 }
 
-func (a *PatientAggregate) UnassignBed(ctx context.Context, currentBedID uuid.NullUUID) error {
-	if !currentBedID.Valid {
-		// Patient cant be unassigned if not assigned
-		return nil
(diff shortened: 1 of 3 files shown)


MaxSchaefer (changes requested):
Shouldnt we rather solve this in the projections? The bed UUID is not a part of the unassigned event, therefore we should not pass this to `UnassignBed`.
Also please add a test in our stories for this behaviour.

PaulKalho:
> Shouldnt we rather solve this in the projections? The bed UUID is not a part of the unassigned event, therefore we should not pass this to `UnassignBed`. Also please add a test in our stories for this behaviour.

As you referred to the logic checking of the `bedID`: In my opinion, logically wrong events should not be published. We, most probably, have these kind of situations elsewhere in the code. I created an issue for that: https://github.com/helpwave/services/issues/854

FoseFx:
> Shouldnt we rather solve this in the projections?

I think this is fine. It's like the handler reacts to its own event and publishes another one on top.

FoseFx (approved):

MaxSchaefer (approved):

PR #850, opened 2024-10-04
restructure dart protos

### LINKED ISSUE #844: Dart Protos  package does not have libs directory
### Describe the bug

The generated Protos for dart don't include libs directory with `common/v1` and `events/v1` which produces errors on the normal tasks endpoints

1. Import the current helpwave proto package from pub.dev
2. use the tasks_service
3. Build fails with file not found

From `helpwave_proto_dart-0.51.0-dcc524f/lib/services/tasks_svc/v1/patient_svc.pb.dart:16:8:`
`Error when reading '.../helpwave_proto_dart-0.51.0-dcc524f/lib/libs/common/v1/conflict.pb.dart'`

### Additional context

_No response_

PaulKalho:
closes #844


MaxSchaefer:
@DasProffi does this structure looks good to you? Whats the last version of the dart package that worked?

DasProffi:
> @DasProffi does this structure looks good to you? 

No, dart requires packages to have all exported components within the `lib/` folder. Therefore the previous version was better than the new change

> Whats the last version of the dart package that worked?
The version without consistency tokens. The issue is that the consistency token grpc files are not withing the `lib/` folder

MaxSchaefer:
> > @DasProffi does this structure looks good to you?
> 
> No, dart requires packages to have all exported components within the `lib/` folder. Therefore the previous version was better than the new change
> 
> > Whats the last version of the dart package that worked?
> 
> The version without consistency tokens. The issue is that the consistency token gRPC files are not within the `lib/` folder and thus the file reference cannot be resolved when using the proto package's tasks files

Okay. Thanks or the clarification. But in current main, there is everything under the `lib` folder. Does this needs to be directly under `lib`?

DasProffi:
> Okay. Thanks or the clarification. But in current main, there is everything under the `lib` folder. Does this needs to be directly under `lib`?

No, subfolders are fine as long as imports are correct. The primary issue is that the `lib/libs` folder was not uploaded to the proto package, if you look at `helpwave_proto_dart 0.51.0-dcc524f`, the `lib/libs` folder is missing which is the root cause of the issues.

FoseFx:
This does not solve the underlying issue.

PR #851, opened 2024-10-04
Evaluate dive validator

### LINKED ISSUE #836: Evaluate "dive" validator
### Describe the chore

[In a review](https://github.com/helpwave/services/pull/817/files#r1765168732), uncertainty on how "dive" works / when it is needed arose. We should evaluate all `repeated` request messages to make sure they are validated as intended.

PaulKalho:
closes #836 

There were only a few places where `validate` had to be added

### DIFF
--- a/services/property-svc/stories/PropertyValueCRUD_test.go
+++ b/services/property-svc/stories/PropertyValueCRUD_test.go
@@ -635,7 +635,7 @@ func TestCreateAttachAddOptionAttachSelectProperty(t *testing.T) {
 				RemoveOptions: nil,
 				UpsertOptions: []*pb.UpdatePropertyRequest_SelectData_SelectOption{
 					{
-						Id:          "",
+						Id:          uuid.NewString(),
 						Name:        hwutil.PtrTo("Option 2"),
 						Description: nil,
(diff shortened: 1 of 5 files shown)


services/property-svc/stories/PropertyValueCRUD_test.go
```diff
@@ -635,7 +635,7 @@ func TestCreateAttachAddOptionAttachSelectProperty(t *testing.T) {
 				RemoveOptions: nil,
 				UpsertOptions: []*pb.UpdatePropertyRequest_SelectData_SelectOption{
 					{
-						Id:          "",
+						Id:          uuid.NewString(),
```
FoseFx, line 638: Nowhere else we require the frontend to provide a would-be uuid for a new resource. I don't think this should be the exception.

FoseFx (changes requested):

services/property-svc/stories/PropertyValueCRUD_test.go
PaulKalho: Thanks, didn't think about it

FoseFx (approved):

PR #858, opened 2024-10-07
chore: increase WaitForProjectionsToSettle time in CI

MaxSchaefer:


MaxSchaefer:
Besides that change, we could implement a retry mechanism for all stories at https://github.com/helpwave/services/blob/main/services/tasks-svc/stories/setup_test.go#L58.

MaxSchaefer:
@FoseFx

FoseFx (approved):

Issue #811, opened 2024-09-11
Let GetUserID and GetOrganizationID panic


FoseFx:
### Describe the chore

Currently, common's GetUserID and GetOrganizationID may return an error, if (and only if) the corresponding values were not attached to the context by the authFunc and handleOrganizationIDForAuthFunc middlewares.

For user-ids this can only happen, when either the service was not Setup() with enabled auth, or a route handler is in skipAuthForMethods, and yet called GetUserID. Either way, this is not an expected runtime error, instead indicates a programming mistake. It makes little sense to **always** handle these. I assume the grpc library will catch the panic and return an error, but we should check this first / implement this anyway.

For organization ids we currently have possible client error to attribute a missing id to. But this will change soon, when we start using jwt-claims for org ids. And as we currently deploy with a fallback org set, this will never error.

Both functions are currently used ~60 times, with error handling this removes at least 180 loc. And likely a lot more, as other functions may only error due to their usage of either one of those functions.

In general: user-error / environment-error -> error, invalid state -> panic

We should prefix them to `Must`, to indicate a possible panic to developers, which should make sure this function makes sense to call in their current context.

FoseFx:
Blocked by #832

MaxSchaefer:
Same for https://github.com/helpwave/services/pull/859#discussion_r1808924731

PR #859, opened 2024-10-10
chore: add updates-svc

### LINKED ISSUE #871: `updates-svc` to provide a realtime updates endpoint for our clients
### Describe the feature

The `updates-svc` should provide a gRPC server-side stream to our clients to give them the ability to listen on our EventStoreDB stream.

We currently have event sourced and non-event sourced entites. Event sourced entites are easy to implement. Non-event sourced entities need to dispatch updates events later on. The latter is not part of this issue.

To enable a certain form of collaboration, the UI must always be up to date. 

### Additional context

_No response_

MaxSchaefer:
Closes #871

Todo before merge
- [x] Deploy to staging


libs/common/auth/auth.go
```diff
@@ -26,9 +27,11 @@ var (
 	oauthConfig             *oauth2.Config
 	Verifier                *oidc.IDTokenVerifier
 	authSetupDone           bool
+	FakeTokenValidFor       = time.Second * 30
 )
 
 type ClaimsKey struct{}
+type TokenExpires struct{}
```
FoseFx, line 34: btw why are these exported? they should not be!

FoseFx, line 34:
```suggestion
type TokenExpiresKey struct{}
```

libs/common/auth/auth.go
```diff
@@ -140,49 +143,50 @@ func (t IDTokenClaims) AsExpected() error {
 // VerifyIDToken verifies the correctness of the accessToken and returns its claim.
 // The claim is checked to be as expected.
 // Service must be set up with auth!
-func VerifyIDToken(ctx context.Context, token string) (*IDTokenClaims, error) {
+func VerifyIDToken(ctx context.Context, token string) (*IDTokenClaims, *time.Time, error) {
```
FoseFx, line 146: document second return parameter

libs/common/auth/auth.go
```diff
@@ -206,6 +210,16 @@ func GetAuthClaims(ctx context.Context) (*IDTokenClaims, error) {
 	}
 }
 
+// SessionValidUntil returns time.Time when the session gets marked as expired
+func SessionValidUntil(ctx context.Context) (time.Time, error) {
+	res, ok := ctx.Value(TokenExpires{}).(*time.Time)
+	if !ok {
+		return time.Now(), status.Error(codes.Internal, "tokenExpires not in context, set up auth")
```
FoseFx, line 217: this is not a valid expected state, thus warrants a panic, not an error

proto/services/updates_svc/v1/updates_svc.proto
```diff
@@ -0,0 +1,44 @@
...
+  string aggregate_id = 3;
+  string aggregate_type = 4;
+}
+
+//
+// ReceiveUpdatesRequest
+//
+
+message ReceiveUpdatesRequest {
+  optional uint64 revision = 1;
```
FoseFx, line 27: document this key

proto/services/updates_svc/v1/updates_svc.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+//
+// ReceiveUpdatesResponse
+//
+
+message ReceiveUpdatesResponse {
+  uint64 revision = 1;
```
FoseFx, line 35: document these keys

FoseFx: what makes an event an entity and what a domain event?

FoseFx: and why would the client care?

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+	return &UpdatesGrpcService{
+		es: es,
+	}
+}
+
+func (s *UpdatesGrpcService) ReceiveUpdates(req *pb.ReceiveUpdatesRequest, stream pb.UpdatesService_ReceiveUpdatesServer) error {
+	ctx, span, log := telemetry.StartSpan(stream.Context(), "updates.ReceiveUpdates")
+	defer span.End()
+
+	tokenExpires, err := auth.SessionValidUntil(ctx)
```
FoseFx, line 28:
```suggestion
	tokenExpiresAt, err := auth.SessionValidUntil(ctx)
```

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+		return err
+	}
+
+	tokenExpiresIn := time.Until(tokenExpires)
+
+	ctx, cancel := context.WithCancel(ctx)
+	// Close stream after session is invalid
+	time.AfterFunc(tokenExpiresIn, func() {
+		log.Debug().Msg("closing stream, token expired")
+		cancel()
```
FoseFx, line 39: will these be cleaned up if the stream breaks?

else I can see the GC holdon to this function, which holds on to the cancel() functions, which holds on to the context

might cause a memory leak?

it returns something you can cancel this with, we should do that (by defer- ing)

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err
+	}
+	defer esStream.Close()
+
+	for {
+		esdbEvent := esStream.Recv()
+
+		select {
+		case <-stream.Context().Done():
```
FoseFx, line 80: isn't this the wrong context? we cancel the ctx, which is a fresh new wrapped context created above.

services/updates-svc/internal/updates/grpc.go
FoseFx: Instead use context.WithTimeout()

services/updates-svc/internal/updates/grpc.go
FoseFx: also add a logline here

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+			return nil
+		default:
+		}
+
+		if esdbEvent.SubscriptionDropped != nil {
+			esStream.Close()
+			log.Debug().Msg("subscription dropped, close stream")
```
FoseFx, line 87: does esdbEvent.SubscriptionDropped supply more information, which may be interesting to be logged?

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+			break
```
FoseFx, line 88:
```suggestion
			return nil
```

services/updates-svc/internal/updates/grpc.go
```diff
@@ -0,0 +1,133 @@
...
+
+		res := &pb.ReceiveUpdatesResponse{
+			Revision: *esdbEvent.EventAppeared.Commit,
+			Event: &pb.ReceiveUpdatesResponse_DomainEvent{
+				DomainEvent: domainEvent,
+			},
+		}
+
+		if err := stream.Send(res); err != nil {
+			log.Error().Err(err).Msg("cannot send on stream")
```
FoseFx, line 127:
```suggestion
			log.Error().Err(err).Msg("cannot send update response on stream")
```

services/updates-svc/stories/setup_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	zlog.Info().
+		Str("eventstoreEndpoint", eventstoreEndpoint).
+		Msg("containers are up")
+
+	// prepare env
+	hwtesting.SetCommonEnv()
+	hwtesting.SetEventstoreEnv(eventstoreEndpoint)
+
+	// TODO: spice migrations (PR #812)
```
FoseFx, line 33:
```suggestion
```

services/updates-svc/stories/updates_test.go
```diff
@@ -0,0 +1,52 @@
...
+	assert.NoError(t, err)
+	assert.NoError(t, stream.CloseSend())
+}
+
+func TestAutoClosingWhenTokenExpiresReceiveUpdateStream(t *testing.T) {
+	ctx := context.Background()
+	updatesClient := updatesServiceClient()
+
+	auth.FakeTokenValidFor = time.Second * 3
+	upperBound := auth.FakeTokenValidFor + (time.Second * 1)
```
FoseFx, line 28: there is time.Round() and time.Truncate()

services/updates-svc/stories/updates_test.go
```diff
@@ -0,0 +1,52 @@
...
+	lowerBound := auth.FakeTokenValidFor - (time.Second * 1)
+
+	req := &pb.ReceiveUpdatesRequest{}
+
+	timeBeforeOpeningStream := time.Now()
+	stream, err := updatesClient.ReceiveUpdates(ctx, req)
+	assert.NoError(t, err)
+
+	_, cancel := context.WithTimeout(ctx, auth.FakeTokenValidFor+(time.Second*3))
```
FoseFx, line 37: why do you create a context that you dont use at all?

FoseFx: contexts are immutable

services/updates-svc/stories/updates_test.go
```diff
@@ -0,0 +1,52 @@
...
+		if err == io.EOF {
+			elapsedTime := time.Since(timeBeforeOpeningStream)
+			if elapsedTime > upperBound || elapsedTime < lowerBound {
+				t.Errorf("Connection timeout is out of bounds (upper bound %s, lower bound %s) with %s", upperBound, lowerBound, elapsedTime)
+				return
+			}
+			return
+		}
+		assert.NoError(t, err)
+	}
```
FoseFx, line 51: what if the stream never EOFs?

libs/common/auth/auth.go
MaxSchaefer: https://github.com/helpwave/services/issues/811#issuecomment-2427021965

services/updates-svc/stories/updates_test.go
MaxSchaefer: Uhiii very nice!

services/updates-svc/stories/updates_test.go
FoseFx: why would it? the context that is being canceled is not used, and nothing else will be triggered that would warrant a failed test

MaxSchaefer: Now that you mention it ... weird that I tested this behaviour before and it worked. _Fixed._

PR #878, opened 2024-10-24
Let GetUserID and GetOrganizationID panic

### LINKED ISSUE #811: Let GetUserID and GetOrganizationID panic

FoseFx:
closes #811

### DIFF
--- a/libs/common/auth/auth.go
+++ b/libs/common/auth/auth.go
@@ -238,22 +238,44 @@ func SessionValidUntil(ctx context.Context) (time.Time, error) {
 	}
 }
 
-func GetUserID(ctx context.Context) (uuid.UUID, error) {
+// MaybeGetUserID can be used instead of GetUserID
+func MaybeGetUserID(ctx context.Context) *uuid.UUID {
 	res, ok := ctx.Value(userIDKey{}).(uuid.UUID)
 	if !ok {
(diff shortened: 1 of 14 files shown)


MaxSchaefer (approved):
No occurrences in `property-svc`?

PR #884, opened 2024-11-04
feat: publishing non event-sourced events

MaxSchaefer:
This pull request enables our event to be defined in proto message. An event message contains the option `libs.events.v1.event_type`. This option gets evaluated by our event logic to resolve the event type directly from the proto defintion.

I also introduce the concept of an EntityEvent. Those events are quickly raised events in non event sourced services.


proto/libs/events/v1/options.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package libs.events.v1;
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "gen/libs/events/v1";
+
+extend google.protobuf.MessageOptions {
+  string event_type = 2123;
```
FoseFx, line 10: explain this number

libs/hwes/event.go
```diff
@@ -48,7 +56,35 @@ type metadata struct {
 	TraceParent string `json:"trace_parent"`
 	// The Timestamp represents the time when the event was created. Using the built-in eventstoreDB timestamp is
 	// discouraged.
-	Timestamp time.Time `json:"timestamp"`
+	Timestamp   time.Time `json:"timestamp"`
+	DataIsProto bool      `json:"data_is_proto"`
+}
+
+// MessageToEventName reflects the passed EventType message extension of the passed message.
```
FoseFx, line 63: this documention raises more questions than it answers

what does it do
then how does it do it, if relevant

libs/hwes/event.go
```diff
@@ -48,7 +56,35 @@ type metadata struct {
...
+// When the EventType is nil, the current default EventType of proto/libs/events/v1 gets used.
```
FoseFx, line 64:
```suggestion
// When the EventType is nil, the current default EventType of proto/libs/events/v1 gets used.
```

instead, drop the second parameter and suffix the function with v1 or alike

libs/hwes/event.go
```diff
@@ -48,7 +56,35 @@ type metadata struct {
...
+func MessageToEventName(protoMessage proto.Message, eventType *protoimpl.ExtensionInfo) string {
```
FoseFx, line 65: "To" implies conversion, but we extract a field here

libs/hwes/event.go
```diff
@@ -104,6 +146,21 @@ func NewEvent(aggregate Aggregate, eventType string, opts ...EventOption) (Event
 	return evt, nil
 }
 
+func NewEventFromProto(aggregate Aggregate, message proto.Message, opts ...EventOption) (Event, error) {
```
FoseFx, line 149:
```suggestion
func FromProto(aggregate Aggregate, message proto.Message, opts ...EventOption) (Event, error) {
```

libs/hwes/eventstoredb/entity_event.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"telemetry"
+
+	"github.com/EventStore/EventStore-Client-Go/v4/esdb"
+	"github.com/google/uuid"
+	"google.golang.org/protobuf/proto"
+	"hwes"
+)
+
+type EntityAggregate struct {
```
FoseFx, line 14: what is an entity aggregate? how does it differ from a normal aggregate? why are they needed?

libs/hwes/eventstoredb/entity_event.go
```diff
@@ -0,0 +1,63 @@
...
+	a := &EntityAggregate{}
+	a.AggregateBase = hwes.NewAggregateBase(hwes.AggregateType(entityType), entityId)
+	return a
+}
+
+func (a *EntityAggregate) GetTypeID() string {
+	return fmt.Sprintf("entity_%s-%s", a.GetType(), a.GetID())
+}
+
+// SaveEntityEvent creates a temporary EntityAggregate to create
```
FoseFx, line 28: not relevant for the devs using the function (i think? or does this cause some side effect?)

services/tasks-svc/internal/ward/ward.go
```diff
@@ -77,14 +86,23 @@ func (s ServiceServer) CreateWard(ctx context.Context, req *pb.CreateWardRequest
 	// add to "recently used"
 	tracking.AddWardToRecentActivity(ctx, wardID.String())
 
+	// emit event
```
FoseFx, line 89:
```suggestion
	// store event
```

proto/services/updates_svc/v1/updates_svc.proto
```diff
@@ -6,7 +6,9 @@ option go_package = "gen/services/updates-svc/v1";
 
 // An event that gets raised on non-event sourced entities
 message EntityEvent {
-  string entity_id = 1;
+  string event_type = 1;
```
FoseFx, line 9: again, why is this split needed? why would a client even care about this impl. detail?

services/updates-svc/stories/updates_test.go
```diff
@@ -24,6 +38,61 @@ func TestOpenAndClosingReceiveUpdatesStream(t *testing.T) {
...
+	bedId := uuid.New()
+
+	req := &pb.ReceiveUpdatesRequest{}
+	stream, err := updatesClient.ReceiveUpdates(ctx, req)
+	require.NoError(t, err)
+
+	go func() {
+		// test receive one event that got emitted after subscribing
+		res, err := stream.Recv()
+		requireTrueBool(t, assert.NoError(t, err))
```
FoseFx, line 56: you can't use require, but can use assert? isn't that largly the same library

services/updates-svc/stories/updates_test.go
```diff
@@ -4,16 +4,30 @@ import (
...
 
+	"github.com/google/uuid"
+
 	"github.com/stretchr/testify/require"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// requireTrueBool because require. cannot be used in Goroutines and to prevent many if-conditions
+func requireTrueBool(t *testing.T, b bool) {
```
FoseFx, line 23:
```suggestion
func requireTrue(t *testing.T, b bool) {
```

services/updates-svc/stories/updates_test.go
```diff
@@ -24,6 +38,61 @@ func TestOpenAndClosingReceiveUpdatesStream(t *testing.T) {
...
+	stream, err := updatesClient.ReceiveUpdates(ctx, req)
+	require.NoError(t, err)
+
+	go func() {
+		// test receive one event that got emitted after subscribing
+		res, err := stream.Recv()
+		requireTrueBool(t, assert.NoError(t, err))
+
+		requireTrueBool(t, assert.NotNil(t, res.GetEntityEvent()))
+		requireTrueBool(t, assert.Equal(t, "bed", res.GetEntityEvent().GetEntityType()))
```
FoseFx, line 59: res.GetEntityEvent() can be a variable

libs/hwes/event.go
MaxSchaefer: We create a new event from the passed proto message. I like the verbosity there. It also indicates a "constructor" like method.

services/updates-svc/stories/updates_test.go
MaxSchaefer: Jup. The underlying .Fail() functions behave differently in Goroutines. require.() calls .FailNow()

proto/services/updates_svc/v1/updates_svc.proto
MaxSchaefer: Any good wording to not use "entity" or "aggregate"?

FoseFx: I don't think I made myself clear.

Why do we, additionally to "normal" Events also have EntityEvents?

MaxSchaefer: To differentiate between non event-sourced and event-sourced entities. But yes, I dont want to show this implementation detail to the client either. But I dont have a good abstraction for the response ... lets strich with only entities ...

FoseFx: why is any abstraction needed at all? just return a normal event

MaxSchaefer: Its not about the event. Its about the entity or aggregate.

FoseFx: yes, just pretend its an aggregate

libs/hwes/eventstoredb/entity_event.go
```diff
@@ -0,0 +1,66 @@
...
+	"telemetry"
+
+	"github.com/EventStore/EventStore-Client-Go/v4/esdb"
+	"github.com/google/uuid"
+	"google.golang.org/protobuf/proto"
+	"hwes"
+)
+
+// Extending entityAggregate to override GetTypeID down below
+type entityAggregate struct {
```
FoseFx, line 15: why do we need this type? why do we not use normal aggregates for everything? then we can switch to an event-sourced approach *really* easily if we want to

FoseFx: why? are the events not the same as when we'd use then to construct projections with them? why would we want to disregard all existing events?

wasn't the point to have the write-part of the ES approach ready to use? we just don't *source* from these events?

MaxSchaefer: _We discussed this via audio_

- [  ] Use aggregate type to differentiate between event sourced and non event sourced event streams

services/tasks-svc/internal/task-template/task_template.go
```diff
@@ -86,13 +102,29 @@ func (ServiceServer) CreateTaskTemplate(
 		Str("taskTemplateID", templateID.String()).
 		Msg("taskTemplate created")
 
+	// store event
+	if err := eventstoredb.SaveEntityEventForAggregate(ctx, s.es, NewTaskTemplateAggregate(templateID),
```
FoseFx, line 106: A cleaner way is to 

- create a new EntityAggregateBase type, which embeds AggregateBase
- EntityAggregateBase then implements a custom Save() method, which does the same as SaveEntityEventForAggregate
- TaskTemplateAggregate, BedAggregate etc embed EntityAggregateBase instead of AggregateBase
- handler call `NewTaskTemplateAggregate(templateID).Save(...)`

libs/hwes/event.go
```diff
@@ -36,6 +41,7 @@ type Event struct {
 	OrganizationID *uuid.UUID
 	// w3c trace context
 	TraceParent string
+	DataIsProto bool
```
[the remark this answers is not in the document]
MaxSchaefer: Yes. We should use mime types instead. But the goal is to quickly move all events in proto and get rid of this bool as well. Let me add a quick comment explaining this.

services/tasks-svc/internal/task-template/task_template.go
MaxSchaefer: Yes. Sounds good. I've created a new issue for this.

FoseFx (approved):

Issue #877, opened 2024-10-24
Permissions


FoseFx:
PropertyService (#812)
- [x] CreateProperty
- [x] GetProperty
- [x] UpdateProperty
- [x] GetProperties

PropertySetService (#880)
- [x] CreatePropertySet
- [x] GetPropertySet

PropertyValueService (#885)
- [x] AttachPropertyValue
- [x] GetAttachedPropertyValues

PropertyViewsService (#885)
- [x] UpdatePropertyViewRule

BedService (#880)
- [x] CreateBed
- [x] GetBed
- [x] GetBeds
- [x] GetBedsByRoom
- [x] GetBedByPatient
- [x] UpdateBed
- [x] DeleteBed

PatientService (#882)
- [x] CreatePatient
- [x] GetPatient
- [x] GetPatientByBed
- [x] GetPatientsByWard
- [x] GetPatientAssignmentByWard
- [x] GetPatientDetails
- [x] GetPatientList
- [x] GetRecentPatients
- [x] UpdatePatient
- [x] AssignBed
- [x] UnassignBed
- [x] DischargePatient
- [x] ReadmitPatient
- [x] DeletePatient

RoomService (#880)
- [x] CreateRoom
- [x] GetRoom
- [x] GetRooms
- [x] UpdateRoom
- [x] DeleteRoom
- [x] GetRoomOverviewsByWard

TaskService (#882)
- [x] CreateTask
- [x] UpdateTask
- [x] GetTask
- [x] GetTasksByPatient
- [x] GetTasksByPatientSortedByStatus
- [x] GetAssignedTasks
- [x] AssignTask
- [x] UnassignTask
- [x] CreateSubtask
- [x] UpdateSubtask
- [x] DeleteSubtask
- [x] RemoveTaskDueDate
- [x] DeleteTask

TaskTemplateService (#885)
- [x] CreateTaskTemplate
- [x] DeleteTaskTemplate
- [x] DeleteTaskTemplateSubTask
- [x] UpdateTaskTemplate
- [x] UpdateTaskTemplateSubTask
- [x] CreateTaskTemplateSubTask
- [x] GetAllTaskTemplates
- [x] GetTaskTemplate

WardService (#880)
- [x] CreateWard
- [x] GetWard
- [x] GetWards
- [x] GetRecentWards
- [x] UpdateWard
- [x] DeleteWard
- [x] GetWardDetails
- [x] GetWardOverviews

OrganizationService
- [x] CreateOrganization
[...]

PR #885, opened 2024-11-04
Another Batch of Permissions

FoseFx:
progresses #877 

- ProperySet
- PropertyValue
- Bed
- TaskTemplates


FoseFx:
WIth Max being OOO, can you take a look at this?

PaulKalho:
> WIth Max being OOO, can you take a look at this?

I will take a look :)

spicedb/_version
```diff
@@ -1 +1 @@
-2
+3
```
FoseFx, line 1: remind me to bump this

services/tasks-svc/internal/bed/bed.go
```diff
@@ -161,6 +160,22 @@ func (ServiceServer) GetBedByPatient(
...
+	// check permissions
+	user := commonPerm.UserFromCtx(ctx)
+	checks := make([]hwauthz.PermissionCheck, 0)
+	checks = append(checks,
+		hwauthz.NewPermissionCheck(user, patientPerm.PatientCanUserGet, patientPerm.Patient(patientId)))
+
+	if result != nil {
+		checks = append(checks,
+			hwauthz.NewPermissionCheck(user, perm.BedCanUserGet, perm.Bed(result.BedID)),
+			hwauthz.NewPermissionCheck(user, roomPerm.RoomCanUserGet, roomPerm.Room(result.RoomID)),
```
PaulKalho, line 172: wouldn't be the check for `RoomCanUserGet` enough, as it implies (through transitivity) that the `bed` can also "be get"?

spicedb/task_template.test.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - "ward:a#organization@organization:a"
+  - "ward:b#organization@organization:b"
+
+  # task_template A belongs to ward A,
+  # and template B to user Bob
+  - "task_template:a#owner@ward:a"
+  - "task_template:b#owner@user:bob"
+
+tests:
+  - name: If the owner is a user, only the owner of a template can RUD it
```
PaulKalho, line 29:
```suggestion
  - name: If the owner is a user, only the owner of a template can GUD it
```

PaulKalho (approved):

services/tasks-svc/internal/bed/bed.go
FoseFx: yes, currently that's true. but in the future we maybe have different roles, some may not have access to certain beds, and thus should not see them here

PR #896, opened 2024-12-03
handle spicedb deletes

FoseFx:
Progresses #877

Keep in mind that this approach will leave leaks (e.g., when deleting an org, all relations to its users and wards will be deleted, but not the relations between a room and such a ward)
We believe that's ok for now.


MaxSchaefer:
> Keep in mind that this approach will leave leaks (e.g., when deleting an org, all relations to its users and wards will be deleted, but not the relations between a room and such a ward)
We believe that's ok for now.
Thanks. LGTM! Please open a follow up issue for this.

MaxSchaefer (approved):

felixevers:
> Member

Documented in #897

PR #898, opened 2024-12-05
refactor: store value changes instead

### LINKED ISSUE #860: Proper model for PropertyValue values
### Describe the chore

Currently, we cast it into an interface{}, which throws away vital information.

Is this a string because it's an option, or a text, or a multi-select option?

We could embrace TypedValue and give it proper serialization to esdb

FoseFx:
Cherry-picked from the now abandoned op. concurrency branch

Closes #860


services/property-svc/internal/property-value/aggregate/aggregate.go
```diff
@@ -72,7 +75,12 @@ func (a *PropertyValueAggregate) onPropertyValueUpdated(evt hwes.Event) error {
 		return err
 	}
 
-	a.PropertyValue.Value = payload.Value
+	if payload.Change.ValueRemoved {
+		a.PropertyValue.Value = nil
+		return nil
+	}
+
+	payload.Change.Apply(a.PropertyValue.Value)
```
MaxSchaefer, line 83: We do not need to re-assign this due to the reference above?

services/property-svc/internal/property-value/projections/property_value_postgres_projection/property_value_postgres_projection.go
```diff
@@ -91,38 +86,28 @@ func (p *Projection) onPropertyValueCreated(ctx context.Context, evt hwes.Event)
...
-			val, ok := payload.Value.(string)
-			if !ok {
-				return errors.New("could not assert string"), hwutil.PtrTo(esdb.NackActionPark)
-			}
-			id, err := uuid.Parse(val)
+		if isSingleSelectChange {
+			id, err := uuid.Parse(*change.SingleSelectValue)
 			if err != nil {
-				return err, hwutil.PtrTo(esdb.NackActionPark)
+				return fmt.Errorf("option change is not a uuid: %w", err), hwutil.PtrTo(esdb.NackActionSkip)
```
MaxSchaefer, line 104: Why skip? We should park and investigate such issues.

MaxSchaefer (changes requested):

services/property-svc/internal/property-value/aggregate/aggregate.go
FoseFx: which reference above?

services/property-svc/internal/property-value/projections/property_value_postgres_projection/property_value_postgres_projection.go
FoseFx: the payload will never be valid, so parking makes little sense

services/property-svc/internal/property-value/aggregate/aggregate.go
MaxSchaefer: Line #67

MaxSchaefer (approved):

services/property-svc/internal/property-value/aggregate/aggregate.go
FoseFx: that's in onPropertyValueCreated, not onPropertyValueUpdated

Issue #865, opened 2024-10-15
golint: fix linter issues and enable more linters


FoseFx:
### Describe the chore

The following linters are disabled, because we still need to fix their issues:

  - [ ] cyclop
  - [ ] dupl
  - [ ] err113
  - [ ] exhaustruct
  - [ ] gci 
  - [ ] gocognit 
  - [ ] goconst 
  - [ ] gocyclo 
  - [ ] godox 
  - [ ] nestif 
  - [ ] nlreturn 
  - [ ] paralleltest
  - [ ] protogetter
  - [ ] revive 
  - [ ] stylecheck 
  - [ ] tagliatelle 
  - [ ] testpackage

PR #913, opened 2024-12-19
chore: paralleltest

FoseFx:
Re: #865

Enables paralleltest linter, by calling `t.Parallel()` for every test, for this:
- remove global state for db and otel Registry (now injected into start-up context)
- add middlewares to propagate these values into the request contexts
- rename GetDB to MustGetDB, and require context as parameter
- add AddUserToOrganization (to unify a common pattern in test code into a proper function)

This should also enable test code to create other services than the own.


felixevers (approved):
LGTM

PR #931, opened 2025-03-03
Remove flyio

felixevers:


.github/workflows/migrations.yaml
```diff
@@ -1,219 +0,0 @@
...
-          services: ${{ steps.changes.outputs.all_changed_files }}
-
-  disallow-modifications:
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - name: Detect modifications in migrations
```
FoseFx, line 58: your over-deleting useful checks here

FoseFx (changes requested):

.github/workflows/migrations.yaml
felixevers: Prbly right I'll dive into this later.

PR #932, opened 2025-05-19
add migrate script to non-standalone image

FoseFx:


Dockerfile.service
```diff
@@ -1,11 +1,45 @@
...
+#
+FROM debian:12.8 AS deps
+
+
+# renovate: datasource=github-releases depName=golang-migrate/migrate
+ARG MIGRATE_VERSION=4.18.1
+
+
+RUN apt update && \
+	apt install -y wget
```
felixevers, line 15: Just a quick note here. Do you mind adding https://github.com/hadolint/hadolint and pinning versions here?

Dockerfile.service
```diff
@@ -1,11 +1,45 @@
...
+#
+
+WORKDIR /build/spice
+
+COPY go.* ./
+COPY libs libs
+COPY spicedb/migrations spicedb/migrations
+RUN go mod download
+COPY cmd/spice cmd/spice
+RUN CGO_ENABLED=0 GOOS=linux go build -a -o spice cmd/spice/spice.go
```
felixevers, line 37: I guess summarize this real quick. One `COPY` one `RUN`.

felixevers (changes requested):

Dockerfile.service
FoseFx: pretty sure this was for dependency caching reasons

Dockerfile.service
FoseFx: either way I'm not touching it right now, no time and will to debug this right now

felixevers (approved):
#933

